PR #1, opened 2015-09-03
Create base Django app with API authetication

marcofucci:


marcofucci:
@SteveMarshall @ibrechin this is basically the same as bank-admin but with related settings / name changed.

ibrechin:
:+1:

Apparently I don't have write access to this repository, though.

marcofucci:
@ibrechin I think you do now

PR #6, opened 2015-09-29
Add shared assets setup

teneightfive:
Assets are now being shared across all 3 frontend app and
are coming from money-to-prisoners-common. This adds the setup needed
to import assets/templates from that package.

This change is part of the following story:

#### User Story [#103077962](https://www.pivotaltracker.com/story/show/103077962)
As a frontend developer,
I want assets/templates used across all frontend apps to be consistent and shared,
So that I don't have to make changes in multiple locations

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.py]
+indent_size = 4
--- a/mtp_prisoner_location_admin/settings/base.py
+++ b/mtp_prisoner_location_admin/settings/base.py
@@ -13,6 +13,14 @@ https://docs.djangoproject.com/en/1.8/ref/settings/
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 import os
 import sys
+import json
+
+from os.path import abspath, join, dirname
+
+here = lambda *x: join(abspath(dirname(__file__)), *x)
+PROJECT_ROOT = here("..")
+root = lambda *x: join(abspath(PROJECT_ROOT), *x)
+bower_dir = lambda *x: join(json.load(open(root('..', '.bowerrc')))['directory'], *x)
 
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 sys.path.insert(0, os.path.join(BASE_DIR, 'apps'))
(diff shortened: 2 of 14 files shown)


mtp_prisoner_location_admin/settings/base.py
```diff
@@ -64,6 +87,13 @@
     },
 ]
 
+
+TEMPLATE_CONTEXT_PROCESSORS = global_settings.TEMPLATE_CONTEXT_PROCESSORS + (
+    'moj_utils.context_processors.debug',
+    'django.core.context_processors.request'
+)
```
ibrechin: Context processors should now be included in the ['OPTIONS']['context_processors'] attribute of the TEMPLATES setting.

The django.templates.context_processors.request entry that is already there replaces django.core.context_processors.request so you can just delete this line and add moj_utils.context_processors.debug to the list above (@line 80).

teneightfive:
retest this please

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
SteveMarshall: Why not trim trailing whitespace on Markdown?

mtp_prisoner_location_admin/settings/base.py
```diff
@@ -23,6 +31,8 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
+TEMPLATE_DEBUG = True
```
SteveMarshall: Do we override this in the production settings? Oh, apparently we don't even do that for `DEBUG`. OK, well, we should probably have a task next sprint for setting up production configs.

.editorconfig
tyom: > When you do want to insert a `<br />` break tag using Markdown, you end a line with two or more spaces, then type return.

– [Markdown syntax](http://daringfireball.net/projects/markdown/syntax#p)

SteveMarshall: God, I'd forgotten about that entirely! Thanks @tyom!

SteveMarshall:
This is applying the changes from ministryofjustice/money-to-prisoners-cashbook#69 to this repo, so marking blocked until that's resolved.

mtp_prisoner_location_admin/settings/base.py
teneightfive: Yep, there are a few things that need to be setup properly for production I think.

Probably best as a separate task.

PR #10, opened 2015-10-06
Allow file upload

ibrechin:
This PR adds an upload form for the submission of prisoner location files.

The exact format of the prisoner location file is still something of an unknown 
as no sample files have been received, but I would suggest that this work be merged
so that work can begin on the frontend.


ibrechin:
@marcofucci take a look when you're ready

mtp_prisoner_location_admin/templates/prisoner_location_admin/dashboard.html
```diff
@@ -1,4 +1,6 @@
 {% extends "base.html" %}
+{% load i18n %}
 
 {% block content %}
+<a href="{% url "location_file_upload" %}">{% trans "Updata Prisoner Locations" %}</a>
```
marcofucci: I think you mean 'Upload'

mtp_prisoner_location_admin/apps/prisoner_location_admin/views.py
```diff
@@ -0,0 +1,24 @@
...
+        kwargs = super(LocationFileUploadView, self).get_form_kwargs()
+        kwargs['request'] = self.request
+        return kwargs
+
+    def form_valid(self, form):
+        try:
+            form.update_locations()
+            return super(LocationFileUploadView, self).form_valid(form)
+        except forms.ValidationError as e:
+            form.add_error(None, e)
```
marcofucci: why do you need this line?

mtp_prisoner_location_admin/apps/prisoner_location_admin/views.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def get_form_kwargs(self):
+        kwargs = super(LocationFileUploadView, self).get_form_kwargs()
+        kwargs['request'] = self.request
+        return kwargs
+
+    def form_valid(self, form):
+        try:
+            form.update_locations()
+            return super(LocationFileUploadView, self).form_valid(form)
```
marcofucci: Maybe you can add a flash message to notify the user that everything went OK?

mtp_prisoner_location_admin/apps/prisoner_location_admin/tests/test_forms.py
```diff
@@ -0,0 +1,43 @@
...
+        file_data, _ = generate_testable_location_data()
+        file_data += '\nA1234GY,1997-9-2'
+
+        request = self.factory.post(
+            reverse('location_file_upload'),
+            {'location_file': get_csv_data_as_file(file_data)}
+        )
+        form = LocationFileUploadForm(request.POST, request.FILES, request=request)
+
+        self.assertFalse(form.is_valid())
```
marcofucci: I'd check the error msg as well, just to be sure that it's that particular validation error being raised and not a different one.

If you don't like copying/pasting the string here maybe you can extend the error_messages on the form field and reference it here.

mtp_prisoner_location_admin/apps/prisoner_location_admin/tests/test_forms.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_location_file_empty_file_invalid(self):
+        request = self.factory.post(
+            reverse('location_file_upload'),
+            {'location_file': get_csv_data_as_file('')}
+        )
+        form = LocationFileUploadForm(request.POST, request.FILES, request=request)
+
+        self.assertFalse(form.is_valid())
```
marcofucci: I'd check the error msg as well, just to be sure that it's that particular validation error being raised and not a different one.

If you don't like copying/pasting the string here maybe you can extend the error_messages on the form field and reference it here.

marcofucci:
@ibrechin a few things in my comments but nothing really major, looks good overall :-)

mtp_prisoner_location_admin/apps/prisoner_location_admin/views.py
ibrechin: Are you referring to the fact that normally raising a ValidationError is caught and converted to a form error automatically? If so, it's because that only happens when it occurs in the 'clean' phase of form processing, whereas this is in the method that is called after the 'clean' phase is completed successfully.

The reason for this is that I didn't want to add validation logic (beyond just the file format) in the client app, so instead any problems with the actual data are found in the API processing and returned as a 400, then reraised as a ValidationError (which is manually caught and displayed as a form error).

mtp_prisoner_location_admin/templates/prisoner_location_admin/dashboard.html
ibrechin: I'll leave that to the front-end devs :)

mtp_prisoner_location_admin/apps/prisoner_location_admin/views.py
marcofucci: makes sense :)

ibrechin:
@marcofucci changed as suggested, ready to merge if you're happy

PR #11, opened 2015-10-06
avoid installing unnecessary packages

maxf:
Don't install browser-sync and other dependent packages
on non-local environments, as those packages are only
useful on developer machines
[updates #104986742]


marcofucci:
@maxf do you mind rebasing from master?

maxf:
@marcofucci done

marcofucci:
@maxf the check fails, I guess something depends on browser-sync when building the assets with gulp

SteveMarshall:
I think this needs the fixes you've applied in ministryofjustice/money-to-prisoners-cashbook#74?

PR #21, opened 2015-10-22
remove the dashboard page

maxf:
... which didn't serve any purpose


test/features/dashboard.feature
```diff
@@ -1,9 +0,0 @@
-Feature: Signing in
-  As a signed-in user
-  I want to be able to access the uploads page from the dashboard
-  So that I can go and upload files
-
-  Scenario: Go to the upload page
-    Given I am signed in
-    And I go to the "Dashboard" page
-    Then I should see a "Update Prisoner Locations" link to "/upload"
```
SteveMarshall: Is there a test for going to the update page instead of the dashboard?

ibrechin: Yeah, possibly the first scenario in upload_file.feature could be changed slightly?

maxf: Added a line to the sign-in tests.

PR #32, opened 2015-11-26
Use hyphens instead of underscores in test usernames

ibrechin:


SteveMarshall:
@ibrechin Looks good to me; can you rebase against master?

PR #64, opened 2016-01-26
Alter prison location file parsing to match known format

ibrechin:
Having received an example file, we can now correct the parsing rules
to handle the actual format of the report.

Details to be dealt with:
  - Header row
  - Trailing non-data rows to be discarded
  - Different column ordering
  - Date format incompatible with API


ibrechin:
retest this please

PR #106, opened 2016-05-23
Add basic security views for payments grouped by sender or prisoner

ushkarev:


mtp_noms_ops/apps/security/tests/test_utils.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+
+
+class UtilTestCase(unittest.TestCase):
+    def test_currency_formatting(self):
+        from security.templatetags.security import currency
```
ibrechin: I would prefer to have all imports at the top of the file.

mtp_noms_ops/apps/security/urls.py
```diff
@@ -3,8 +3,23 @@
...
 
-
+security_test = user_test(required_permissions)
 urlpatterns = [
-    url(r'^$', user_test(required_permissions)(views.SecurityDashboardView.as_view()),
+    url(r'^$', security_test(views.dashboard_view),
         name='security_dashboard'),
+
+    url(r'^sender-grouped/$', security_test(views.SenderGroupedView.as_view()),
+        name='security_sender_grouped'),
```
ibrechin: Would it be easier to use Django's in-built URL namespacing rather than manually prefixing all these URL names with ```security_```?

PR #118, opened 2016-06-14
Provide navigation between prisoner location upload and security views

ushkarev:
…for users who can access both


maxf:
Maybe a couple of functional tests should be added?

PR #127, opened 2016-06-15
Tweaks to security credits lists

ushkarev:


ushkarev:
@maxf this one's ready for you now :)

mtp_noms_ops/assets-src/javascripts/modules/security-forms.js
```diff
@@ -56,6 +56,9 @@ exports.SecurityForms = {
 
         $creditList.find('caption').remove();
         $creditList.find('.CollapsingTable').removeClass('CollapsingTable');
+        $creditList.find('tr th:first-of-type, tr td:first-of-type').each(function() {
+          $(this).remove();
+        });
         $thisRow.find('td').each(function () {
```
maxf: Does this work in IE8? first-of-type isn't supported

ushkarev: i figured jquery would deal with that, it doesn't use native selector mechanisms does it?

maxf: We should test anyway

PR #130, opened 2016-06-17
Improve print layout of security pages by hiding form

ushkarev:


mtp_noms_ops/apps/security/forms.py
```diff
@@ -77,8 +79,8 @@ class SecurityForm(GARequestErrorReportingMixin, forms.Form):
     """
     page = forms.IntegerField(min_value=1)
     page_size = 20
-    received_at_0 = forms.DateField(label=_('Received since'), help_text=_('eg 01/06/2016'), required=False)
-    received_at_1 = forms.DateField(label=_('Received before'), help_text=_('eg 01/06/2016'), required=False)
+    received_at_0 = forms.DateField(label=_('Received from'), help_text=_('eg 01/06/2016'), required=False)
+    received_at_1 = forms.DateField(label=_('Received to'), help_text=_('eg 01/06/2016'), required=False)
```
maxf: Shouldn't we have just one 'Received', and below From and To, like on other forms?

PR #150, opened 2016-07-08
Sort group results by sender or prisoner

maxf:
Part of https://www.pivotaltracker.com/story/show/121503857


maxf:
@ushkarev @ibrechin leaving this one for you to solve, plz:
`dictsort` freaks out when one of the keys is None, apparently. I'm not sure what the best way around is.

PR #155, opened 2016-07-12
Slightly prettier print API errors when location file upload fails

ibrechin:
They were previously a big mess of braces and parentheses.


ibrechin:
retest this please

PR #170, opened 2016-08-15
Add ability to download transaction search results as csv file

ibrechin:


mtp_noms_ops/apps/security/views.py
```diff
@@ -195,3 +197,26 @@ def form_valid(self, form):
...
         return render(self.request, self.results_template_name, context)
+
+
+class CreditsExportView(SecurityView):
+    form_class = CreditsForm
+
+    def form_valid(self, form):
+        data = form.get_complete_object_list()
+        csvdata = export_as_csv(data)
+        response = HttpResponse(csvdata, content_type='text/plain')
```
ushkarev: why not `text/csv`, how does IE behave?

PR #191, opened 2016-11-10
Upload prisoner locations asynchronously using uwsgi spooler

ibrechin:
If the application is running in a uwsgi context, then when a user
uploads a prisoner location file it is saved a job is spooled to
upload it to the backend API in chunks. If this occurs, the user
who uploaded the file will be sent an email notification if there
are any problems encountered when uploading the file.


ushkarev:
@ibrechin `UploadTests.test_upload_valid_file` fails now!

PR #207, opened 2016-12-08
Implement new security home page

vickymg:


ushkarev:
i would say this is not quite ready to merge. for one thing, the links in the proposition header depend on your access rights.

PR #257, opened 2017-03-07
Add support for retrieving a prisoner photo from NOMIS

ibrechin:
This photo is to be used on the prisoner profile page.


ibrechin:
@vickymg this is the backend work for the photo. On the page you should be able to check if the API integration is available with `{% if nomis_api_available %}` and if the photo is present with `{% if photo %}` (`photo` is b64 encoded image data so should be able to use it with just ``.

mtp_noms_ops/templates/security/prisoners-detail.html
```diff
@@ -2,20 +2,32 @@
...
-      {{ view.title }}
       {% if prisoner %}
+        {% if nomis_api_available %}
+          <div class="mtp-prisoner-image">
+            {% if photo %}
+              <img src="{{ photo }}" alt="{% trans 'Photo of the prisoner' %}"/>
+            {% else %}
+              <img src="{% static 'images/placeholder-image.png' %}"
+              srcset="{% static 'images/placeholder-image@2x.png' %} 2x"
+              alt="{% trans 'Placeholder image for prisoner photo' %}" />
```
ushkarev: this wouldn't be a placeholder so screen readers would say the wrong thing

mtp_noms_ops/templates/security/prisoners-detail.html
```diff
@@ -2,20 +2,32 @@
...
   <header>
     {% language_switch %}
-    {% include 'security/includes/save-search.html' %}
     <h1 class="heading-xlarge">
-      {{ view.title }}
       {% if prisoner %}
+        {% if nomis_api_available %}
+          <div class="mtp-prisoner-image">
+            {% if photo %}
+              <img src="{{ photo }}" alt="{% trans 'Photo of the prisoner' %}"/>
```
ushkarev: should be `data:image/jpeg;base64,{{ photo }}` i think. but also depends on how NOMIS encodes the image – sometimes base64 encoded files include line breaks that we’d have to strip out.

ushkarev:
oh and a rebase is needed now

PR #293, opened 2017-08-09
Force asynchronous creation of credit export that's emailed to the user

ushkarev:
…when there are lots of credits and it'd likely time out


mtp_noms_ops/templates/security/includes/export-credits.html
```diff
@@ -0,0 +1,28 @@
+{% load i18n %}
+{% load mtp_common %}
+
+{% if form.total_count <= 5000 %}
```
ibrechin: Could see this value in settings.

PR #297, opened 2017-08-15
Fix filter bugs

ibrechin:


mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -410,11 +413,26 @@ def test_displays_results(self, mocked_connection):
         self.assertIn('Debit card', response_content)
 
 
+@contextmanager
+def temp_file(data):
+    path = '/tmp/export.xlsx'
```
ushkarev: perhaps [tempfile](https://docs.python.org/3/library/tempfile.html) would be more elegant..?

mtp_noms_ops/apps/security/tasks.py
```diff
@@ -24,26 +25,25 @@ def zip_credit_export(attachment_name, output):
 
 
 @spoolable(body_params=('user', 'session', 'filters'))
-def email_credit_csv(*, user, session, endpoint_path, filters, export_description,
-                     attachment_name, zip_attachment=True):
+def email_credit_xlsx(*, user, session, endpoint_path, filters, export_description,
+                      attachment_name, zip_attachment=True):
```
ushkarev: since XLSX is already a zip file, it makes sense to remove zipping..?

ibrechin: Yeah, I was thinking about that, but it does make some small difference (like 3%), and for really large files that might be a significant number of bytes. Probably only relevant if someone does actually try and download the entire database though...

mtp_noms_ops/apps/security/tests/test_views.py
ibrechin: Sure

PR #298, opened 2017-08-22
Implement redesigned filter tabs

ibrechin:


ibrechin:
Should clear filters display all the time?

ushkarev:
@ibrechin i think this is good now. shall we check with flo re the clear button?

PR #324, opened 2018-05-21
Load cell location on prisoner details page after page load

ibrechin:
As requests to the NOMIS API can be quite slow, load the cell
location after page load using AJAX.


mtp_noms_ops/apps/security/views.py
```diff
@@ -291,3 +279,21 @@ def prisoner_image_view(request, prisoner_number):
...
+                levels = {
+                    level['type']: level['value'] for level in
+                    location['housing_location']['levels']
+                }
+                response_data['housing_location'] = '-'.join([
+                    levels['Wing'], levels['Landing'], levels['Cell']
+                ])
+        except RequestException:
+            logger.warning('Could not load location for %s' % prisoner_number)
+    return JsonResponse(response_data)
```
ushkarev: how do you feel about some caching? like with photos: `patch_cache_control(response, private=True, max_age=???)`

PR #327, opened 2018-06-13
Allow for prisoner location data to have variable level types

ushkarev:


mtp_noms_ops/apps/security/views.py
```diff
@@ -287,13 +287,13 @@ def prisoner_nomis_info_view(request, prisoner_number):
...
-                }
-                response_data['housing_location'] = '-'.join([
-                    levels['Wing'], levels['Landing'], levels['Cell']
-                ])
+                housing_location = location['housing_location']
+                if 'levels' in housing_location:
+                    response_data['housing_location'] = '-'.join(
+                        level['value'] for level in housing_location['levels']
+                    )
+                elif 'description' in housing_location:
```
ibrechin: I don't think this condition will ever get hit, as a description is always split into levels in common.

PR #333, opened 2018-08-31
HMPPS employee confirmation

ushkarev:


mtp_noms_ops/apps/security/views.py
```diff
@@ -300,3 +307,55 @@ def prisoner_nomis_info_view(request, prisoner_number):
...
+                self.success_url = success_url
+            return super().form_valid(form)
+        else:
+            self.save_user(hmpps_non_employee_flag, True)
+            self.request.session.flush()
+            return self.non_employee()
+
+    def non_employee(self):
+        self.title = _('You can’t use this tool')
+        self.template_name = 'hmpps-non-employee.html'
```
ibrechin: This seems weird. Why not just add a `TemplateView`?

PR #336, opened 2018-10-26
Limit exporting objects even by email

ushkarev:


ushkarev:
@ibrechin you ok with the form/view rearrangement? they were getting a little unwieldy…

PR #342, opened 2018-12-07
Create Google analytics events for the search parameters used

ibrechin:
This is to allow us to track what filters are used by users, as well
as when they try and export or print data.

### DIFF
--- a/mtp_noms_ops/assets-src/javascripts/modules/form-analytics.js
+++ b/mtp_noms_ops/assets-src/javascripts/modules/form-analytics.js
@@ -1,0 +1,37 @@
+'use strict';
+
+var analytics = require('analytics');
+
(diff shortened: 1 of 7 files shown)


mtp_noms_ops/assets-src/javascripts/modules/form-analytics.js
```diff
@@ -0,0 +1,61 @@
...
+      'event', {
+        eventCategory: 'SecurityForms',
+        eventAction: 'print',
+        eventLabel: this.$print.href()
+      }
+    );
+  },
+
+  onSubmit: function (e) {
+    var $inputs = this.$form.find('input');
```
ushkarev: there's also `select`… i'd recommend using `this.$form.serialize()`

PR #364, opened 2019-03-22
Change GA event sending to be compatible with blanket param stripping

ibrechin:
Send event attributes as positional arguments rather than a dict allows
a final dict argument to override page, location and title.


ibrechin:
retest this please

ibrechin:
retest this please

PR #387, opened 2019-07-08
Prison switcher

marcofucci:
This adds a new Prison Switcher area to the top of some pages showing the first 4 prisons the user has set in his/her settings with a link to the exiting page responsible for changing this if needed.

If more than 4 prisons exist in the user settings, a 'x more' text is shown as well.

The user is redirected back to the page he/she was coming from after saving the form via the prison switcher.

I added the section to the new `security/base.html` template as it only applies to security pages but that makes it a bit weird as the dashboard now extends `security/base.html`. This dashboard template should really be in the `security` folder but it was already in the root folder so I'm okay to keep it this way.
@ushkarev let me know if you have anything else in mind instead.

Screenshot 2019-07-08 at 16 21 39

### DIFF
--- a/mtp_noms_ops/apps/security/context_processors.py
+++ b/mtp_noms_ops/apps/security/context_processors.py
@@ -1,5 +1,7 @@
 from urllib.parse import urlencode
 
+from django.contrib.auth import REDIRECT_FIELD_NAME
+
 from .utils import (
     can_see_notifications, can_choose_prisons, is_nomis_api_configured
 )
@@ -28,3 +30,12 @@ def initial_params(request):
         ('prison', prison['nomis_id'])
         for prison in request.user_prisons
     ], doseq=True)}
+
+
+def common(_):
+    """
+    Context Processor for common / core logic, e.g. making some variable available in the templates.
+    """
+    return {
+        'REDIRECT_FIELD_NAME': REDIRECT_FIELD_NAME,
+    }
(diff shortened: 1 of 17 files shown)


ushkarev:
since the prison switcher is intended to show you that the results on the page are limited to that prison set, i don't think it should show on these pages:
- mtp_noms_ops/templates/security/sender.html (because the credits there are not filtered)
- mtp_noms_ops/templates/security/prisoner.html (because the credits and disbursements there are not filtered)
- mtp_noms_ops/templates/security/credit.html (because that credit might be for any prison)
- mtp_noms_ops/templates/security/disbursement.html (because that disbursement might be for any prison)

ushkarev:
oh and forgot to mention, we've a convention to prefix classes with `mtp-` so that it's clear that it's our style, not an imported one.

marcofucci:
> since the prison switcher is intended to show you that the results on the page are limited to that prison set, i don't think it should show on these pages:
> 
> * mtp_noms_ops/templates/security/sender.html (because the credits there are not filtered)
> * mtp_noms_ops/templates/security/prisoner.html (because the credits and disbursements there are not filtered)
> * mtp_noms_ops/templates/security/credit.html (because that credit might be for any prison)
> * mtp_noms_ops/templates/security/disbursement.html (because that disbursement might be for any prison)

I didn't realise it was related to the actual search.
I've changed this, it will only appear on the following pages:
- senders.html
- prisoners.html
- disbursements.html
- credits.html

I don't think it should appear on the other pages (e.g. dashboard, review etc.) but let me know if I'm wrong.

marcofucci:
@ushkarev this PR is ready to be reviewed again.
Sorry that I had to force-push but I got the changes from master and added the feature flag check.
It was cleaner with a rebase.

PR #389, opened 2019-07-12
Simple search form for payment sources

marcofucci:
This:

- Adds the ability to search for payment sources using a single input text by matching sender name or email
- Adds a `SEARCH_V2` feature flag so that the new and the old logic can coexist
- refactors the generic security views to that the new logic can be accommodated

Depends on https://github.com/ministryofjustice/money-to-prisoners-api/pull/520


mtp_noms_ops/assets-src/stylesheets/views/_security-form-fields.scss
```diff
@@ -93,3 +93,35 @@ select[disabled] {
     }
   }
 }
+
+#filter-senders {
+  .column-one-third {
+    .mtp-info-box {
+      padding: 1em;
+
```
ushkarev: can you put in a `@include media(mobile) {...}` rule to add a margin; it touches the search button otherwise

mtp_noms_ops/templates/security/senders_list.html
```diff
@@ -0,0 +1,127 @@
...
+        <div class="column-one-third print-hidden">
+          <div class="mtp-info-box">
+            <div class="mtp-info-box__heading">
+              {% trans 'About payment source data' %}
+            </div>
+            <p>
+              {% trans 'One person using many cards displays as many payment sources.' %}
+            </p>
+            <p>
+              {% trans "We can't check names match their cards, we can't match people by name to different cards and we can't check if email addresses are real." %}
```
ushkarev: a little style note, we should always use curly quotation marks and apostrophes even when we're copying other text where this was forgotten. can you pls make it `can’t` not `can't`

mtp_noms_ops/apps/security/forms/object_list.py
```diff
@@ -146,6 +158,29 @@ def get_query_data(self, allow_parameter_manipulation=True):
...
+        required=False,
+        help_text=_('Common or incomplete names may show many results'),
+    )
+
+    # NB: ensure that these templates are HTML-safe
+    filtered_description_template = 'Results containing {filter_description}, ordered by {ordering_description}.'
+    unfiltered_description_template = 'Payment sources, ordered by {ordering_description}'
+
+    description_templates = (
+        ('prisoner number or name "{search}"',),
```
ushkarev: i think this ought to be `“{search}”` for consistency with other quotation marks

mtp_noms_ops/apps/security/forms/object_list.py
```diff
@@ -146,6 +158,29 @@ def get_query_data(self, allow_parameter_manipulation=True):
...
+    Search Form for Senders V2.
+    """
+    search = forms.CharField(
+        label=_('Search payment source name or email address'),
+        required=False,
+        help_text=_('Common or incomplete names may show many results'),
+    )
+
+    # NB: ensure that these templates are HTML-safe
+    filtered_description_template = 'Results containing {filter_description}, ordered by {ordering_description}.'
```
ushkarev: we'll need to check this with the team. the content-reviewed prototypes do not mention ordering. our options are:
1) get a content review to ok this text
2) not mention ordering _at all_
3) have 2 descriptions! one short form for the visible text, the longer-form including ordering for the invisible table caption [probably my preferred option, though it adds complexity]

mtp_noms_ops/apps/security/views/object_base.py
```diff
@@ -112,35 +139,66 @@ def form_valid(self, form):
...
+            'simple_search_form_submitted_input_name': SIMPLE_SEARCH_FORM_SUBMITTED_INPUT_NAME,
+            'is_search_results': self.view_type == ViewType.search_results,
+        }
+
+    def redirect_to_referral_view(self, form):
+        """
+        Returns an HttpResponseRedirect to the referral_view preserving the same kwargs and query string.
+        """
+        view = reverse(self.referral_view, kwargs=self.kwargs)
+        return redirect(f'{view}?{self.request.GET.urlencode()}')
```
ushkarev: quick question, why us this using GET parameters rather than `form.query_string` like before? just simplicity or is there something in GET that's not in the form?

mtp_noms_ops/assets-src/stylesheets/views/_security-form-fields.scss
```diff
@@ -93,3 +93,35 @@ select[disabled] {
...
+        margin-bottom: 0;
+        font-size: 16px;
+      }
+    }
+  }
+}
+
+.mtp-security-form--simple {
+  #id_search-label {
+    font-size: 27px;
```
ushkarev: just an fyi: gds fonts shrink on smaller screens. this label will remain large because of the absolute size. in this case, i'd leave it as you have done. it still looks perfectly fine on mobile

mtp_noms_ops/apps/security/forms/object_list.py
marcofucci: 👍 you mean `‘{search}’` I guess.

mtp_noms_ops/apps/security/views/object_base.py
marcofucci: that's because if the form is invalid query_string won't have the correct params, it's better to preserve the invalid params and use request.GET instead I think.

mtp_noms_ops/apps/security/forms/object_list.py
marcofucci: Ok `“{search}”` is the version that the team is happy with so I'll go with that.

mtp_noms_ops/assets-src/stylesheets/views/_security-form-fields.scss
marcofucci: 👍 didn't like it either but the prototype didn't have the margin, I think Gavin wouldn't mind so I'll add it.

mtp_noms_ops/apps/security/forms/object_list.py
marcofucci: We decided not to show the ordering at all for the time being.

PR #392, opened 2019-07-16
Redesign results list for payment sources

marcofucci:
This adds a new template called `senders_list.html` with a new version of the payment source results list.

~The template is not currently used by any views at the moment but will be.
To check locally, change `security.views.object_list.SenderListView.template_name` to `'security/senders_list.html'`~
I've rebased so the new search now uses this updated template.

**A couple of things to note**
~ I've decided to vertically align the table cells in the middle as it looks tidier especially with those sorting background images. I could change this though.~
- I've made the `Payment method` column not sortable for now as changing it is not straightforward but I could try harder if it needs be

**Before**
Screenshot 2019-07-16 at 13 15 02

**After**
Screenshot 2019-07-23 at 16 25 10

### DIFF
--- a/mtp_noms_ops/templates/security/senders_list.html
+++ b/mtp_noms_ops/templates/security/senders_list.html
@@ -37,7 +37,7 @@
     </div>
 
     {% if form.is_valid %}
-      <div class="mtp-results-list">
+      <div class="mtp-results-list-v2">
         <div class="print-hidden mtp-links--no-panel">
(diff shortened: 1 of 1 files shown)


marcofucci:
@ushkarev Gavin prefers not to add vertical alignment so I need to do a bit more work on this.

marcofucci:
@ushkarev made the changes and ready to be reviewed now

PR #410, opened 2019-08-15
Add query string value mine for current user's prisons

marcofucci:
This adds a new special value `mine` to the list of allowed choices for the query param `prison` in search forms V2.

The value is then automatically expanded into the current user's prisons when making API calls.


marcofucci:
Superseded by #412

PR #412, opened 2019-08-19
Prison selector in search forms V2

marcofucci:
This:
- adds a `prison_selector` field along with the existing `prison` field. Valid options for prison selector are: `All prisons`, `"My prisons"`, or `A specific prison`. When the last option is specified the prison field is required.
- some constants and functions in tests were renamed to keep things consistent
- the existing hide/reveal javascript was renamed to Legacy so that it'll be easier to find and remove when v2 goes live

I would suggest reviewing the commits individually.

Not in this PR and still to be implemented:
- removal of prison switcher in advanced search form and search results page
- extra `prison` column in search results table when using advanced search

Screenshot 2019-08-19 at 14 47 44

### DIFF
--- a/mtp_noms_ops/apps/security/forms/object_list.py
+++ b/mtp_noms_ops/apps/security/forms/object_list.py
@@ -25,6 +25,8 @@ from security.utils import (
     sender_profile_name,
 )
 
+YOUR_PRISONS_QUERY_STRING_VALUE = 'mine'
+
 END_DATE_BEFORE_START_DATE_ERROR_MSG = _('Must be after the start date.')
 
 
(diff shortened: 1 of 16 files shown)


mtp_noms_ops/apps/security/forms/object_list.py
```diff
@@ -47,6 +49,86 @@ def get_api_request_params(self):
...
+
+    when prison_selector == exact a `prison` value must be specified
+    otherwise the form isn't valid.
+    when prison_selector != exact, any `prison` value is reset as not applicable
+    """
+    PRISON_SELECTOR_EXACT_PRISON_CHOICE_VALUE = 'exact'
+    PRISON_SELECTOR_ALL_PRISONS_CHOICE_VALUE = 'all'
+
+    prison_selector = forms.ChoiceField(
+        label=_('Prison selector'),
```
ushkarev: does this visually appear anywhere? presumably error summary box? maybe it ought to be something like 'Choose a prison'… can wait for Jane to decide tho

mtp_noms_ops/templates/security/forms/prison-selector-fieldset.html
```diff
@@ -0,0 +1,51 @@
...
+              {{ label }}
+            {% endif %}
+          </label>
+
+        </div>
+        {% endif %}
+      {% endfor %}
+
+      {% with sub_field=form.prison %}
+      <div class="mtp-prison-selection-row" id="row_{{ sub_field.name }}">
```
ushkarev: what are `div.mtp-prison-selection-row` or `div.mtp-prison-selection-row__prison` used for?

mtp_noms_ops/apps/security/forms/object_list.py
marcofucci: > presumably error summary box?

Yes and _Choose a prison_ is much better, thanks!

mtp_noms_ops/templates/security/forms/prison-selector-fieldset.html
marcofucci: > what are div.mtp-prison-selection-row or div.mtp-prison-selection-row__prison used for?

The result of a copy/paste, I'll get rid of them, thanks!

PR #432, opened 2019-09-24
Add support for cloud platform migration

marcofucci:
This adds the cloud platform migration middleware and sets up the related settings values as env vars.

There's one javascript error when the site is in maintenance or migration mode as `django` is not defined because of the missing `js-i18n.js`.

I'm not too worried about it as we don't really need any javascript (except ga) but I'm not sure if they would trigger sentry errors?

Fixing that is a bit painful but not too difficult and it probably requires changing https://github.com/ministryofjustice/money-to-prisoners-common/blob/master/mtp_common/assets-src/javascripts/modules/date-picker.js#L199 to support non-django-defined cases as well.

@ushkarev what do you think?

Update: fixed the data-picker.js to only init language strings when needed.

### DIFF
--- a/mtp_noms_ops/apps/security/context_processors.py
+++ b/mtp_noms_ops/apps/security/context_processors.py
@@ -28,7 +28,7 @@ def initial_params(request):
             ),
         }
 
-    if not request.user_prisons:
+    if not getattr(request, 'user_prisons', None):
         return {}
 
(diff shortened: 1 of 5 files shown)


marcofucci:
> might be worth experimenting with the middleware being below `django.middleware.locale.LocaleMiddleware`..?

👍 I will move it down.

PR #435, opened 2019-10-14
"Simple search - all prisons" link

marcofucci:
This adds a new link just before the footer on the search results page which, when clicked, extends the same simple search to all prisons.

The link is only shown:
- when using the simple search
- if the search term is != ''
- If the default prisons of the logged-in user != 'all'

This also rewords the search description and adds _from all prisons_ when using the all prisons link above.

Finally, clicks on this new link are saved in GA so that we can check how many people actually use it.

### Simple search results

**Before**
Screenshot 2019-10-14 at 14 53 07

**After**
Screenshot 2019-10-14 at 14 47 05

Note the updated copy as well (that contain the... or the...).

### When clicking on the all prisons link

Screenshot 2019-10-14 at 14 47 26

Note _"from all prisons"_ in the description, _"- all prisons"_ in the header, the lack of the prison switcher, and the prison column.


mtp_noms_ops/templates/security/credits_list.html
```diff
@@ -94,9 +94,19 @@
...
+    <p>
     {% with search_form_item=breadcrumbs|slice:":-1"|last %}
       <a href="{{ search_form_item.url }}" class="button">{% trans 'Search again' %}</a>
     {% endwith%}
+    </p>
+
+    {% if all_prisons_simple_search_link %}
+    <p>
+      <a href="{{ all_prisons_simple_search_link }}">
+        {{ form.search_description.description }} {% trans 'from all prisons' %}
```
ushkarev: when we come to look at jira story MTP-1081, i think it might be worth moving the "from all prisons" bit into the mechanism that generates descriptions in form classes

marcofucci: I agree, I changed that line and added the extra text to the generated description but it might need more refactoring when starting story 1081.

marcofucci:
retest this please

PR #438, opened 2019-10-25
Remove search v1 code

marcofucci:
This gets rid of all search v1 related code.

Entry point to legacy searches will be redirected to search v2. 
Analytics shows that there aren't many but stil some so we should be able to delete them in a few months.

This also gets rid of some legacy javascript events.


marcofucci:
I'll come back ;-)

PR #444, opened 2019-11-13
Add payment started time to credit history

ushkarev:
which is the time a user was on the website (i.e. at ip address)


ushkarev:
retest this please

PR #455, opened 2019-11-28
Accept/Reject credits

marcofucci:
This:
- implements the accept check flow
- implements the reject check flow
- renames payments pending to credits pending

### List
Screenshot 2019-11-28 at 12 36 42

### Accept
Screenshot 2019-11-28 at 12 37 39

### Reject
Screenshot 2019-11-28 at 12 38 14

### DIFF
--- a/mtp_noms_ops/templates/security/checks_list.html
+++ b/mtp_noms_ops/templates/security/checks_list.html
@@ -52,7 +52,7 @@
           {{ check.credit.amount|currency }}
           </td>
           <td class="print-hidden">
-            <a href="" class="button">{% trans 'Accept' %}</a>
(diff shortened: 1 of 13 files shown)


mtp_noms_ops/templates/security/checks_list.html
```diff
@@ -13,11 +13,11 @@ <h1 class="heading-xlarge">{{ view.title }}</h1>
 
   {% include 'mtp_common/includes/message_box.html' %}
 
-  <p>
+  <p class="font-large">
```
ushkarev: i think this should be `lede` not `font-large`, but i'm not concerned if it looks the same. anyway, this part will likely be iterated a bit in the next few weeks

PR #474, opened 2020-02-18
Add nvmrc and missing watchdog dependency

mattbee:
Simple update, adding some missing dependencies that stall 

### Questions

- Should the nvmrc file be added to a deploy ignore config or is it fine as it should have no effect on anything ever? 
- Is watchdog a dev dependency? Any idea off the top of your head what it's used for?


requirements/dev.txt
```diff
@@ -2,3 +2,4 @@
 -r base.txt
 
 money-to-prisoners-common[testing]>=9.16,<9.17
+watchdog>=0.9.0,<1
```
ushkarev: i don't get why this isn't installed automatically by mtp-common

ushkarev:
watchdog is used when you use `./run.py serve` locally to rebuild static assets and refresh the browser when files are updated

requirements/dev.txt
mattbee: 🤷‍♀ me neither, but working on the assumption that this will be skipped if previous installations are successful, then doesn't hurt.

PR #475, opened 2020-02-19
Add Job Information Form

georgeschena:
## Questions/Notes

- ~I'm not happy with the early return in the Job Information `form_valid` method (see inline comment for details).~
- The templates in mtp-common didn't quite fit the requirements of the prototype/design - but we decided it wasn't worth updating them to be able to just include the templates for radios and textarea, a better plan is probably update them and make them more flexible when reworking them in the GDS styles update work.

### Add Job info form for collecting user research data

Jira: https://dsdmoj.atlassian.net/browse/MTP-1266

This commit adds a form to the Intelligence tool which will collect
information about an user's job title, the area of the prison estate
they work in and what their job title involves.

By using the flags used elsewhere in this app we only redirect a user
to a form if they have not previously completed the information.

We considered setting the user flag in the API but that would mean the
flag text had to be set in 2 places and given other flags are set in
this app when user actions occur (choosing prisons for instance) we
decided that this was the right choice.

The custom validation also ensures that if the user selects 'Other' form
the job title list, they still have to enter a job title to be able to
submit the form.

### DIFF
--- a/mtp_noms_ops/apps/security/tests/test_views.py
+++ b/mtp_noms_ops/apps/security/tests/test_views.py
@@ -25,6 +25,7 @@ from security import (
     hmpps_employee_flag,
     not_hmpps_employee_flag,
     required_permissions,
+    provided_job_info_flag,
 )
 from security.forms.object_list import PrisonSelectorSearchFormMixin, PRISON_SELECTOR_USER_PRISONS_CHOICE_VALUE
 from security.models import EmailNotifications
@@ -83,6 +84,19 @@ def no_saved_searches(rsps=None):
     )
 
 
+def mock_post_for_job_info_endpoint(rsps=None):
+    rsps = rsps or responses
+    rsps.add(
+        rsps.POST,
+        api_url('/job-information/'),
+        json={
+            'job_title': 'Evidence collator',
+            'prison_estate': 'National',
+            'tasks': 'Inmate Processing'
+        },
+    )
+
+
 @override_settings(
     CACHES={'default': {'BACKEND': 'django.core.cache.backends.dummy.DummyCache'}},
 )
@@ -133,7 +147,7 @@ class SecurityBaseTestCase(SimpleTestCase):
         email='sam@mtp.local',
         permissions=required_permissions,
         prisons=(SAMPLE_PRISONS[1],),
-        flags=(hmpps_employee_flag, confirmed_prisons_flag,),
+        flags=(hmpps_employee_flag, confirmed_prisons_flag, provided_job_info_flag),
         roles=('security',),
     ):
         return {
@@ -275,6 +289,115 @@ class PrisonSwitcherTestCase(SecurityBaseTestCase):
         )
 
 
+class JobInformationViewTestCase(SecurityBaseTestCase):
+    protected_views = [
+        'security:credit_list',
+        'security:dashboard',
+        'security:disbursement_list',
+        'security:prisoner_list',
+        'security:sender_list',
+    ]
+
+    @responses.activate
+    def test_redirects_when_provided_job_info_flag_is_missing(self):
+        self.login(
+            user_data=self.get_user_data(
+                flags=[
+                    confirmed_prisons_flag,
+                    hmpps_employee_flag,
+                ],
+            ),
+        )
+        for view in self.protected_views:
+            response = self.client.get(reverse(view), follow=True)
+            self.assertContains(response, '<!-- job_information -->')
+
+    @responses.activate
+    def test_can_view_app_when_user_has_provided_job_info_flag(self):
+        self.login(
+            user_data=self.get_user_data(
+                flags=[
+                    confirmed_prisons_flag,
+                    hmpps_employee_flag,
+                    provided_job_info_flag,
+                ]
+            ),
+        )
+        response = self.client.get(reverse('security:dashboard'), follow=True)
+        self.assertContains(response, '<!-- security:dashboard -->')
+
+    @responses.activate
+    def test_successful_form_post(self):
+        mock_post_for_job_info_endpoint()
+        self.login(user_data=self.get_user_data(
+            flags=[confirmed_prisons_flag, hmpps_employee_flag])
+        )
+        responses.add(
+            responses.PUT,
+            api_url('/users/shall/flags/%s/' % provided_job_info_flag),
+            json={}
+        )
+
+        response = self.client.post(reverse('job_information'), data={
+            'job_title': 'Evidence collator',
+            'prison_estate': 'National',
+            'tasks': 'yas sir'
+        }, follow=True)
+
+        self.assertContains(response, '<!-- security:dashboard -->')
+
+    @responses.activate
+    def test_unsuccessful_form_post(self):
+        mock_post_for_job_info_endpoint()
+        self.login(
+            user_data=self.get_user_data(
+                flags=[
+                    confirmed_prisons_flag,
+                    hmpps_employee_flag
+                ]
+            )
+        )
+        responses.add(
+            responses.PUT,
+            api_url('/users/shall/flags/%s/' % provided_job_info_flag),
+            json={}
+        )
+
+        response = self.client.post(reverse('job_information'), data={
+            'job_title': '',
+            'prison_estate': '',
+            'tasks': ''
+        }, follow=True)
+
+        self.assertContains(response, '<!-- job_information -->')
+        self.assertContains(response, 'This field is required')
+
+    @responses.activate
+    def test_form_adds_provided_job_info_flag(self):
+        mock_post_for_job_info_endpoint()
+        self.login(
+            user_data=self.get_user_data(
+                flags=[
+                    confirmed_prisons_flag,
+                    hmpps_employee_flag
+                ]
+            )
+        )
+        responses.add(
+            responses.PUT,
+            api_url('/users/shall/flags/%s/' % provided_job_info_flag),
+            json={}
+        )
+
+        response = self.client.post(reverse('job_information'), data={
+            'job_title': 'Evidence collator',
+            'prison_estate': 'National',
+            'tasks': 'yas sir'
+        }, follow=True)
+
+        self.assertIn(provided_job_info_flag, response.context['user'].user_data['flags'])
+
+
 class HMPPSEmployeeTestCase(SecurityBaseTestCase):
     protected_views = [
         'security:credit_list',
@@ -319,6 +442,7 @@ class HMPPSEmployeeTestCase(SecurityBaseTestCase):
                 flags=[
                     confirmed_prisons_flag,
                     hmpps_employee_flag,
+                    provided_job_info_flag,
                 ]
             )
         )
@@ -334,7 +458,7 @@ class HMPPSEmployeeTestCase(SecurityBaseTestCase):
     @responses.activate
     def test_employee_flag_set(self):
         self.login(user_data=self.get_user_data(
-            flags=['abc', confirmed_prisons_flag])
+            flags=['abc', confirmed_prisons_flag, provided_job_info_flag])
         )
         responses.add(
             responses.PUT,
@@ -354,6 +478,7 @@ class HMPPSEmployeeTestCase(SecurityBaseTestCase):
             user_data=self.get_user_data(
                 flags=[
                     confirmed_prisons_flag,
+                    provided_job_info_flag,
                 ],
             ),
         )
@@ -2100,7 +2225,7 @@ class PrisonerDetailViewTestCase(SecurityViewTestCase):
 class NotificationsTestCase(SecurityBaseTestCase):
     def login(self, rsps):
         super().login(user_data=self.get_user_data(flags=[
-            hmpps_employee_flag, confirmed_prisons_flag,
+            hmpps_employee_flag, confirmed_prisons_flag, provided_job_info_flag,
         ]), rsps=rsps)
 
     def test_no_notifications_not_monitoring(self):
@@ -2393,6 +2518,7 @@ class SettingsTestCase(SecurityBaseTestCase):
         with responses.RequestsMock() as rsps:
             self.login(user_data=self.get_user_data(flags=[
                 hmpps_employee_flag, confirmed_prisons_flag,
+                provided_job_info_flag,
             ]), rsps=rsps)
             rsps.add(
                 rsps.GET,
@@ -2422,6 +2548,7 @@ class SettingsTestCase(SecurityBaseTestCase):
         with responses.RequestsMock() as rsps:
             self.login(user_data=self.get_user_data(flags=[
                 hmpps_employee_flag, confirmed_prisons_flag,
+                provided_job_info_flag,
             ]), rsps=rsps)
             rsps.add(
                 rsps.GET,
--- a/mtp_noms_ops/apps/security/urls.py
+++ b/mtp_noms_ops/apps/security/urls.py
@@ -8,10 +8,19 @@ from mtp_common.auth.api_client import get_api_session
 from mtp_noms_ops.utils import user_test
 from security import required_permissions, views
 from security.searches import get_saved_searches, populate_new_result_counts
-from security.utils import can_manage_security_checks, can_skip_confirming_prisons, is_hmpps_employee
+from security.utils import (
+    can_manage_security_checks,
+    can_skip_confirming_prisons,
+    is_hmpps_employee,
+    has_provided_job_information,
+)
 
 
 def security_test(view, extra_tests=None):
+    view = user_passes_test(
+        has_provided_job_information,
+        login_url='job_information',
+    )(view)
     view = user_passes_test(
         can_skip_confirming_prisons,
         login_url='confirm_prisons',
--- a/mtp_noms_ops/apps/security/utils.py
+++ b/mtp_noms_ops/apps/security/utils.py
@@ -179,7 +179,8 @@ def has_provided_job_information(user):
 
 
 def can_skip_confirming_prisons(user):
-    already_confirmed = confirmed_prisons_flag in user.user_data.get('flags', [])
+    flags = user.user_data.get('flags') or []
+    already_confirmed = confirmed_prisons_flag in flags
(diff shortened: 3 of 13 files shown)


mtp_noms_ops/apps/settings/views.py
```diff
@@ -152,3 +152,38 @@ def get_context_data(self, **kwargs):
...
+    def get_success_url(self):
+        if 'next' in self.request.GET:
+            return '{path}?{query}'.format(
+                path=self.success_url,
+                query=urlencode({'next': self.request.GET['next']})
+            )
+        return self.success_url
+
+    def form_valid(self, *args, **kwargs):
+        if has_provided_job_information(self.request.user):
```
mattbee: I'm not overly happy about this early return, but it handles and edge case of submitting the form if it happens to be open in 2 browsers and submitted twice accidentally (getting a HTTP 500 error from the API).

I couldn't see how the choose prison handled this however, although it doesn't seem to have the same issue. Suggestions are welcome.

ushkarev: that's quite an unlikely edge case i think. also, i have no issues with early returns personally

mtp_noms_ops/assets-src/stylesheets/views/_job-information-form.scss
```diff
@@ -0,0 +1,11 @@
+.mtp-job-information-form {
+    .mtp-conditional-wrapper {
+        @extend .panel;
+        border-left-width: 5px;
+        margin-left: 18px;
+        padding: 0 0 0 26px;
+    }
+    [name='tasks'] {
```
ushkarev: i'd use the id of the element in the selector, i think it's less esoteric :)

mtp_noms_ops/assets-src/stylesheets/views/_job-information-form.scss
```diff
@@ -0,0 +1,11 @@
+.mtp-job-information-form {
+    .mtp-conditional-wrapper {
+        @extend .panel;
```
ushkarev: hopefully we can rip this out once we move noms-ops to the gds design system. this is duplication of `.mtp-security-advanced-search .mtp-conditional-wrapper`

mtp_noms_ops/apps/settings/views.py
```diff
@@ -152,3 +152,38 @@ def get_context_data(self, **kwargs):
...
+            return redirect(self.get_success_url())
+
+        if self.form_class(self.request.POST).is_valid():
```
ushkarev: this is a little weird. the method's signature is actually `form_valid(self, form)` so you already have a _valid_ form variable to work with

mtp_noms_ops/apps/settings/views.py
```diff
@@ -152,3 +152,38 @@ def get_context_data(self, **kwargs):
...
+            session = get_api_session(self.request)
+            if self.request.POST['job_title'] == 'Other' and self.request.POST['other_title'] != '':
```
ushkarev: you should be reading values from the validated form variable not the raw POST data

ushkarev: also, i'd have this logic on the form class. e.g. the clean method could add another key to the cleaned_data dictionary

mtp_noms_ops/apps/settings/views.py
```diff
@@ -152,3 +152,38 @@ def get_context_data(self, **kwargs):
...
+
+
+class JobInformationView(FormView):
+    title = _('Help us improve this service')
+    template_name = 'settings/job-information.html'
+    success_url = reverse_lazy('security:dashboard')
+    form_class = JobInformationForm
+
+    def get_success_url(self):
+        if 'next' in self.request.GET:
```
ushkarev: does this get populated and does it actually do something? also, if we allow a redirection to happen, we need to validate the target url. c.f. [eligibility view](https://github.com/ministryofjustice/money-to-prisoners-noms-ops/blob/master/mtp_noms_ops/apps/security/views/eligibility.py#L42)

mtp_noms_ops/apps/settings/views.py
```diff
@@ -152,3 +152,38 @@ def get_context_data(self, **kwargs):
...
+                job_title = self.request.POST['job_title']
+
+            session.post('/job-information/', json={'title': job_title,
+                                                    'prison_estate': self.request.POST['prison_estate'],
+                                                    'tasks': self.request.POST['tasks']})
+
+            save_user_flags(self.request, provided_job_info_flag)
+            return redirect(self.get_success_url())
+        else:
+            self.form_class(self.request.POST)
```
ushkarev: this line has no effect

mtp_noms_ops/templates/settings/job-information.html
```diff
@@ -0,0 +1,97 @@
...
+                                </div>
+                                {% if value == 'Other' %}
+                                <div class="mtp-conditional-wrapper" id="mtp-conditional-job_title-Other">
+                                {% include 'mtp_common/forms/field.html' with field=form.other_title only %}
+                                </div>
+                                {% endif %}
+                            {% endfor %}
+                        {% endwith %}
+                    </div>
+                    </fieldsetclass="mtp-conditional-container">
```
ushkarev: ohoh

mtp_noms_ops/templates/settings/job-information.html
```diff
@@ -0,0 +1,97 @@
+{% extends 'base.html' %}
```
ushkarev: HTML, CSS and JS use a 2-space indent style and i think it's worth indenting these files the same as the rest of the app

mtp_noms_ops/apps/security/urls.py
```diff
@@ -394,5 +403,5 @@ def dashboard_view(request):
             RedirectView.as_view(pattern_name='security:prisoner_list'),
         ),
         name='prisoner_list_legacy',
-    ),
+    )
```
ushkarev: i disagree :) i think a trailing comma keeps blocks symmetrical so re-ordering and adding new ones makes the diff clearer (i.e. no lines are modified, you only get new lines or removed lines)

mtp_noms_ops/templates/settings/job-information.html
```diff
@@ -0,0 +1,97 @@
...
+    {% include 'mtp_common/includes/message_box.html' %}
+
+    <div class="grid-row">
+        <div class="column-two-thirds">
+
+            <form method="post" class="mtp-job-information-form">
+                {% csrf_token %}
+
+                {% if form.errors %}
+                {% include 'mtp_common/forms/error-summary.html' with form=form only %}
```
ushkarev: personally, i prefer indenting django template tags in the same way as html elements so that you can see where the nesting happens

mtp_noms_ops/apps/settings/forms.py
```diff
@@ -264,3 +264,38 @@ def get_confirmation_query_string(self):
         query_dict = self.request.GET.copy()
         query_dict['prisons'] = prisons
         return urlencode(query_dict, doseq=True)
+
+
+class JobInformationForm(ApiForm):
+    job_titles = [
+        ('Evidence collator', 'Evidence collator'),
```
ushkarev: the values should be wrapped as lazy translations. even though we might never turn on Welsh, it'll be so much more arduous to put them in later than at the time a feature is worked on

ushkarev: also, the values have spaces, but in your template you have things like `{{ field.html_name }}-{{ value }}` which would be invalid HTML ids

mtp_noms_ops/apps/settings/forms.py
```diff
@@ -264,3 +264,38 @@ def get_confirmation_query_string(self):
...
+    ]
+    job_title = forms.ChoiceField(label=_('What is your job title?'), choices=job_titles,
+                                  widget=forms.RadioSelect, help_text=_('Select one option.'))
+    prison_estate = forms.ChoiceField(label=_('Which area of the prison estate do you work in?'),
+                                      choices=prison_estates,
+                                      widget=forms.RadioSelect, help_text=_('Select one option.'))
+    tasks = forms.CharField(label=_('What are your main tasks?'), widget=forms.Textarea,
+                            help_text=_('Give a brief description.'))
+
+    other_title = forms.CharField(max_length=100, label='Tell us your job title', required=False)
```
ushkarev: there's a missing lazy translation thing here

mtp_noms_ops/apps/settings/views.py
```diff
@@ -152,3 +152,38 @@ def get_context_data(self, **kwargs):
...
+                job_title = self.request.POST['other_title']
+            else:
+                job_title = self.request.POST['job_title']
+
+            session.post('/job-information/', json={'title': job_title,
+                                                    'prison_estate': self.request.POST['prison_estate'],
+                                                    'tasks': self.request.POST['tasks']})
+
+            save_user_flags(self.request, provided_job_info_flag)
+            return redirect(self.get_success_url())
```
ushkarev: maybe `return super(). form_valid(…)` is better but i'd check what that does (as i recall, it's literally this)

mtp_noms_ops/templates/settings/job-information.html
```diff
@@ -0,0 +1,97 @@
...
+                            <button type="submit" class="button">
+                                {% trans 'Continue' %}
+                            </button>
+                        </div>
+                    </form>
+
+                </div>
+            </div>
+
+        {% endblock %}
```
ushkarev: is this the end of `{% block inner_content %}`? it's much more deeply indented here, so i wonder if there are missing closed elements or something iffy going on with the indentation further up

mtp_noms_ops/templates/settings/job-information.html
mattbee: Indeed.

mtp_noms_ops/apps/security/utils.py
```diff
@@ -173,6 +173,11 @@ def is_hmpps_employee(user):
     return hmpps_employee_flag in flags
 
 
+def has_provided_job_information(user):
+    flags = user.user_data.get('flags') or []
```
ushkarev: i like this more than `user.user_data.get('flags', [])`

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -275,6 +289,104 @@ def test_sees_all_prisons(self):
...
+                    hmpps_employee_flag,
+                    provided_job_info_flag,
+                ]
+            ),
+        )
+        response = self.client.get(reverse('security:dashboard'), follow=True)
+        self.assertContains(response, '<!-- security:dashboard -->')
+
+    @responses.activate
+    def test_successfull_form_post(self):
```
ushkarev: i'm being a bit silly, but it's "successful"

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -275,6 +289,104 @@ def test_sees_all_prisons(self):
...
+            json={}
+        )
+
+        response = self.client.post(reverse('job_information'), data={
+            'job_title': '',
+            'prison_estate': '',
+            'tasks': ''
+        }, follow=True)
+
+        self.assertContains(response, 'This field is required')
```
ushkarev: it's worth checking that we're on the right page cos that's a fairly generic error message and could _feasibly_ appear elsewhere

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -275,6 +289,104 @@ def test_sees_all_prisons(self):
...
+            'prison_estate': 'National',
+            'tasks': 'yas sir'
+        }, follow=True)
+
+        self.assertContains(response, '<!-- security:dashboard -->')
+
+    @responses.activate
+    def test_unsuccessfull_form_post(self):
+        mock_post_for_job_info_endpoint()
+        self.login(user_data=self.get_user_data(
```
ushkarev: i'm slightly surprised that the mixture of a hanging indent and a normal one didn't cause flake8 to complain. i think it's best to stick to one style in a single expression

mtp_noms_ops/assets-src/stylesheets/views/_job-information-form.scss
mattbee: Yep, it could have been dropped in there but thought it better to keep this separate (doing the same for the JS looked more problematic though).

mtp_noms_ops/apps/settings/views.py
mattbee: It does something now!

mtp_noms_ops/apps/settings/views.py
mattbee: Yep, this is another hangover from initially working with a `post` method

mtp_noms_ops/apps/security/urls.py
mattbee: I assure you this wasn't a conscious choice on my part ;) I prefer trailing commas everywhere too

mtp_noms_ops/apps/settings/forms.py
mattbee: Good spot - hadn't really thought about the knock on effects of user friendly form values.

mtp_noms_ops/apps/security/utils.py
mattbee: Cool, I find it more readable but consistency is as important, to keep it consistent with the use of default arguments only a few lines below (line 182 of this file), I actually think it might be better using the `default` argument and refactor the other uses in one go if you do like the more verbose syntax better?

What do you think?

ushkarev: i'm 110% for consistency. so the pedant in me would change other occurrences to `user.user_data.get('flags') or []`

mtp_noms_ops/apps/settings/views.py
mattbee: Good plan, this makes it all simpler (at least no more complicated in the clean method and simpler in the view)

mtp_noms_ops/apps/settings/forms.py
```diff
@@ -264,3 +264,47 @@ def get_confirmation_query_string(self):
         query_dict = self.request.GET.copy()
         query_dict['prisons'] = prisons
         return urlencode(query_dict, doseq=True)
+
+
+class JobInformationForm(ApiForm):
+    job_titles = [
+        (_('Evidence collator'), 'Evidence collator'),
```
ushkarev: aren't choice supposed to be value-label pairs not label-value?

mtp_noms_ops/apps/settings/views.py
```diff
@@ -152,3 +152,34 @@ def get_context_data(self, **kwargs):
...
+    template_name = 'settings/job-information.html'
+    form_class = JobInformationForm
+
+    def get_success_url(self):
+        if REDIRECT_FIELD_NAME in self.request.GET:
+            next_page = self.request.GET[REDIRECT_FIELD_NAME]
+            url_is_safe = is_safe_url(
+                url=next_page,
+                allowed_hosts=self.get_success_url_allowed_hosts(),
+                require_https=self.request.is_secure(),
```
ushkarev: this looks like a sensible solution. i think in other places we've used `is_safe_url` https is always required which means it's impossible to test and demo locally

mtp_noms_ops/apps/settings/forms.py
mattbee: You're right - I think I got this from the only place that doesn't label them correctly, but this is a different context so possible correct too (https://github.com/ministryofjustice/money-to-prisoners-noms-ops/blob/6105043aa8a2ee4c0d2acf7b6a4b326b521c0ab9/mtp_noms_ops/apps/security/forms/object_base.py#L89-L90)

ushkarev: the `.name` and `.value` come from the Enum class but we're using the name as the vale and the value as the label in the choices :)

mattbee: I see, yep, lacking context there - sorting that now though

mtp_noms_ops/apps/settings/views.py
mattbee: I assume we've got some SSL enforcement in the cloud platform set up though so this would report correctly as enforced SSL in production?

ushkarev: yeah you cannot access stuff in cloud platform without using ssl

mtp_noms_ops/apps/settings/forms.py
```diff
@@ -264,3 +264,47 @@ def get_confirmation_query_string(self):
...
+class JobInformationForm(ApiForm):
+    job_titles = [
+        ('Evidence collator', _('Evidence collator')),
+        ('Intelligence analyst', _('Intelligence analyst')),
+        ('Intelligence officer', _('Intelligence officer')),
+        ('Intelligence researcher', _('Intelligence researcher')),
+        ('Intelligence support officer', _('Intelligence support officer')),
+        ('Safety analyst', _('Safety analyst')),
+        ('Security analyst', _('Security analyst')),
+        ('Other'_('Other'))
```
ushkarev: there's a little typo in this line

mtp_noms_ops/apps/settings/forms.py
```diff
@@ -264,3 +264,47 @@ def get_confirmation_query_string(self):
...
+        ('National', _('National')),
+    ]
+    job_title = forms.ChoiceField(label=_('What is your job title?'),
+                                  choices=job_titles,
+                                  widget=forms.RadioSelect)
+    prison_estate = forms.ChoiceField(label=_('Which area of the prison estate do you work in?'),
+                                      choices=prison_estates,
+                                      widget=forms.RadioSelect)
+    tasks = forms.CharField(label=_('What are your main tasks?'),
+                            widget=forms.Textarea,
```
ushkarev: as it happens, the `widget` argument won't have any effect because of how our html templates work. no harm having it tho

mattbee: If it's not used, I'd rather remove it,  to avoid confusion and an assumption it is used, good spot.

PR #477, opened 2020-03-12
Create credit review page

mattbee:
Mainly looking for feedback on any erroneous code, probably stuff not needed or that have an easier implementation.


mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -74,17 +75,38 @@ def get_object_list(self):
...
+
+    non_pending_error_msg = "You cannot action this credit as it's not in pending"
 
     def __init__(self, object_id, request, **kwargs):
         super().__init__(**kwargs)
         self.object_id = object_id
+        self.need_attention_date = self.get_need_attention_date()
         self.request = request
 
+    def get_need_attention_date(self):
```
ushkarev: this is now a duplicate of the same method in `CheckListForm`, right? probably makes sense to extract it into a separate function that both forms can use. there'd be less duplication to manage in the future

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -74,17 +75,38 @@ def get_object_list(self):
...
     """
-    non_pending_error_msg = NotImplementedError
+    URGENT_IF_OLDER_THAN = datetime.timedelta(days=3)
+    decision_reason = forms.CharField(
+        label=gettext_lazy('Give details (details are optional when accepting)'),
+        required=False,
+    )
+    fiu_action = forms.CharField(max_length=10)
+
+    non_pending_error_msg = "You cannot action this credit as it's not in pending"
```
ushkarev: this is now only used in one place, right? we might as well put the text there. also, looks like we had forgotten to wrap it in a `gettext` thing in the past

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -93,6 +115,7 @@ def get_object(self):
         """
         try:
             obj = self.session.get(self.get_object_endpoint_path()).json()
+            obj['needs_attention'] = parse_datetime(obj['credit']['started_at']) < self.need_attention_date
             return convert_date_fields(obj, include_nested=True)
```
ushkarev: the `convert_date_fields` parses the `started_at` value, so you could just add the `needs_attention` flag after

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -112,64 +135,43 @@ def clean(self):
...
+                endpoint,
                 json={
-                    'rejection_reason': self.cleaned_data['rejection_reason'],
+                    'decision_reason': self.cleaned_data['decision_reason'],
                 }
             )
             return True
         except RequestException:
-            logger.exception(f'Check {self.object_id} could not be rejected')
+            logger.exception(f'Check {self.object_id} could not be accepted')
```
ushkarev: …could not be *actioned*…?

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -61,48 +61,30 @@ def get_context_data(self, **kwargs):
...
-    form_class = RejectCheckForm
-
-    def form_valid(self, form):
-        if self.request.method == 'POST':
-            result = form.reject()
             if not result:
                 return self.form_invalid(form)
 
+            if form.cleaned_data['fiu_action'] == 'accept':
+                ui_message = 'Credit accepted'
```
ushkarev: you need the `gettext` wrapper where the string is actually defined or else the translation tools can't pick out the message

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -17,7 +17,7 @@ class CheckListView(SecurityView):
     form_class = CheckListForm
 
 
-class ActionCheckView(FormView):
+class BaseAcceptOrRejectCheckView(FormView):
```
ushkarev: is there only one subclass now? could merge into one if so?

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -2737,10 +2737,31 @@ def test_calculation_of_date_before_which_checks_need_attention(self):
...
+                    'count': 1,
+                    'results': [self.SAMPLE_CHECK],
+                },
+            )
+
+            response = self.client.get(reverse('security:check_list'), follow=True)
 
-class AcceptCheckViewTestCase(BaseCheckViewTestCase):
+            content = response.content.decode()
+            self.assertIn('Review <span class="visually-hidden">credit to John Doe</span>', content)
```
ushkarev: tiny fyi: there's a slightly simpler `assertContains` which does the content decoding already by itself

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -2790,15 +2811,15 @@ def test_get(self):
...
+            url = reverse('security:resolve_check', kwargs={'check_id': check_id})
             response = self.client.get(url, follow=True)
 
             self.assertContains(response, 'Accept credit')
             self.assertContains(response, '1234******987 &nbsp; 02/20')
 
     def test_accept_check(self):
         """
-        Test that if one tries to acceot an already rejected check, a validation error is displayed.
+        Test that if one tries to accept an already rejected check, a validation error is displayed.
```
ushkarev: :)

mtp_noms_ops/apps/security/forms/check.py
mattbee: The `gettext_lazy` was called where the variable was used so that's OK, but yeah we can just use the text in place of the variable now 👍

mtp_noms_ops/apps/security/forms/check.py
mattbee: I started but didn't finish that actually. Will see how it goes now. Main problem I had was where to put it that was sensible.

mtp_noms_ops/apps/security/views/check.py
mattbee: Yep, sorted. I misread a title reassignment, which was actually just for the breadcrumbs

mtp_noms_ops/apps/security/forms/check.py
mattbee: I've seen the pattern I used in other Rest Framework projects I looked at, but keen on your thoughts on whether I'm matching your or other contributors patterns for sticking reusable functions before classes that use them - or is there a better place for it and import it for use?

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -112,64 +104,43 @@ def clean(self):
         """
         Makes sure that the check is in pending.
         """
+        status = self.cleaned_data['fiu_action']
+        if 'decision_reason' in self.cleaned_data:
+            reason = self.cleaned_data['decision_reason']
+        else:
+            reason = ''
+
+        if reason == '' and status == 'reject':
```
Staberinde, line 113: If `self.cleaned_data['decision_reason'] = None` it would get through this check when it shouldn't. I would change the first clause of this if statement to test for truthiness rather than testing for a value of empty string. Having said that, this is probably ok as long as `decision_reason` stays a `Charfield` and no-one sets `null=True` in its constructor https://docs.djangoproject.com/en/2.0/ref/models/fields/#null

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -112,64 +104,43 @@ def clean(self):
...
+            msg = forms.ValidationError('This field is required')
```
Staberinde, line 114: So on the surface this looks deeply wrong to me, to instantiate a base class of exception outside of the context of a `raise` statement, but it turns out this is explicitly what the django docs recommend https://docs.djangoproject.com/en/2.0/ref/forms/validation/#raising-multiple-errors and this is even the way the `add_error` method works under the hood https://github.com/django/django/blob/master/django/forms/forms.py#L326 Since it's being coerced within `add_error` you technically don't need to instantiate it here, given that we haven't done so elsewhere, but I really don't mind if you want to leave it as is.

mtp_noms_ops/apps/security/forms/check.py
mattbee: Cheers, would `bool(reason) == False` be how you would check for Falsy values in this context? And sufficient to catch a None value?

Staberinde: Just `if not reason and ...` will do nicely. `None` will evaluate to falsy so that will be covered by this

mtp_noms_ops/apps/security/forms/check.py
mattbee: I think I borrowed this from either somewhere in the code base or django docs, can maybe think about improving later but will ship for now. I can't remember - more reason for more rapid work/deploying :)

PR #483, opened 2020-06-05
Filter out irrelevant demo security checks

ushkarev:
Before delayed capture functionality was enabled, there was a period during which checks were being created and automatically accepted as a preview for FIU. These should not appear in the history page since nobody "actioned" them.
[#MTP-1447](https://dsdmoj.atlassian.net/browse/MTP-1447)


ushkarev:
@mattbee Hopefully this changes nothing significant to prevent the other PRs currently in progress from rebasing nicely!

PR #484, opened 2020-06-08
Normalise security check pages' appearance

ushkarev:
cosmetic changes to make the 3 security check pages look more alike: tables with checks have the same info in the same places and are styled in the same way

### DIFF
--- a/mtp_noms_ops/templates/security/accept_or_reject_check.html
+++ b/mtp_noms_ops/templates/security/accept_or_reject_check.html
@@ -122,11 +122,10 @@
               {{ credit.amount|currency }}
             </td>
             <td>
-              {% if credit.security_check.status == 'accepted' %}
-                <p class="mtp-accepted-tag">{{ credit.security_check.status |upper }}</p>
-              {% else %}
-                <p class="mtp-rejected-tag">{{ credit.security_check.status |upper }}</p>
-              {% endif %}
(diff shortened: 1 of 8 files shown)


mtp_noms_ops/templates/security/accept_or_reject_check.html
```diff
@@ -18,6 +18,7 @@ <h1 class="heading-xlarge">{{ view.title }}</h1>
     {% include 'mtp_common/forms/error-summary.html' with form=form only %}
 
     <h2 class="heading-medium">{% trans 'Credit details' %}</h2>
+    <div class="mtp-results-list table-font-xsmall mtp-check-list">
     <table class="mtp-table mtp-table--review">
```
mattbee, line 22: Is it worth indenting the table now there's a surrounding div, really minor point of course!

ushkarev: yes i totally would! i just didn't want to change every line and have this not be mergeable with #479. actually i checked this morning and the SCSS file doesn't merge. so i don't mind this PR going on top of the other once that's merged. or vice versa if you like?

Staberinde: My strong preference would be that this PR gets merged after #479

PR #487, opened 2020-06-19
Improve security check resolution form for non-pending credits

ushkarev:
There is currently nothing to prevent the view from showing a resolved security check. This would be confusing to users who could click back by accident and try to action it again.
The form would've presented an error, but might as well tell them up-front what the status is. It also allows for potentially using this view as a detail page to link from the decision history.

• hide accept/reject form if already resolved
• show details of decision instead

[MTP-1455](https://dsdmoj.atlassian.net/browse/MTP-1455)


mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -3156,77 +3201,6 @@ def test_accept_check(self):
...
-            response = self.client.post(
-                url,
-                data={
-                    'fiu_action': 'accept',
-                },
-                follow=True
-            )
-
-            self.assertContains(response, 'You cannot action this credit')
-
```
mattbee, line 3229: Is this test definitely redundant? 

Is the possible case where 2 users have the form open at the same time and then one actions the check before the other something that would still trigger this error occurring (i.e has the code doing the outcome of this test also been removed?)

ushkarev: it's redundant in that there is a line-by-line identical test already: `test_invalid_if_check_not_in_pending`. previously, these were on two separate views so made sense, but now that it's one form, they do exactly the same thing. i.e. `test_invalid_if_check_not_in_pending` already tests for the situation you mention

mattbee: Ok, cool, missed the duplicate test in the PR but that's cool then

ushkarev:
sounds like the team is ok to go ahead with this

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -112,20 +110,16 @@ def __init__(self, object_id, request, **kwargs):
...
         try:
             obj = self.session.get(self.get_object_endpoint_path()).json()
             convert_dates_obj = convert_date_fields(obj, include_nested=True)
             convert_dates_obj['needs_attention'] = convert_dates_obj['credit']['started_at'] < self.need_attention_date
             return obj
-        except HttpNotFoundError:
-            self.add_error(None, _('Not found'))
-            return None
         except RequestException:
-            self.add_error(None, _('This service is currently unavailable'))
```
Staberinde, line 127: Would we not want to display an error to the user in this case?

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -3035,6 +3033,53 @@ def test_get(self):
...
+            rsps.add(
+                rsps.GET,
+                api_url(f'/security/checks/{check_id}/'),
+                json=self.SENDER_CHECK_REJECTED
+            )
+            rsps.add(
+                rsps.GET,
+                urljoin(
+                    settings.API_URL,
+                    '/senders/{sender_profile_id}/credits/'.format(
```
Staberinde, line 3054: I've not used the Sentry `request-mock` library (i've tended to use this one instead https://pypi.org/project/requests-mock/), is it agnostic to query strings by default?

mtp_noms_ops/apps/security/forms/check.py
ushkarev: we would but we cannot, because you can only use `add_error` after a form has been "cleaned"

mtp_noms_ops/apps/security/tests/test_views.py
ushkarev: it's a little odd. if you include a query string, it's required to match, if you don't then it doesn't matter. in this test, i just wanted to simplify it and not bother with query string matching as the intention of this test is to check a different thing

mtp_noms_ops/apps/security/forms/check.py
ushkarev: if this method returns None (some kind of error situation), the raises a 404 error. not identical behaviour (different message seen by user), but the line i've removed actually doesn't work.

mtp_noms_ops/apps/security/tests/test_views.py
Staberinde: That's good to know, otherwise i'm sure that would have tripped me up at some point

Staberinde: My only thought is that it might confuse people that the query string is present here but not elsewhere. Might be worth adding https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/487#discussion_r446127320 (or something to that effect) as a code comment in the test maybe?

PR #489, opened 2020-07-03
MTP-1452 Add assign functionality for security checks

mattbee:
## Questions

1. Is this approach a decent thing? Wasn't sure about just having redirects as the handling of the view
2. Will this handle invalid forms? I can't think of a case where the form would be invalid, but guess there could be things I missed
3. Are the tests comprehensive enough - I feel like I have nothing connecting actually hitting the button on the AcceptRejectView

Jira: https://dsdmoj.atlassian.net/browse/MTP-1452

This commit adds the ability to assign a check to a user so that they
can manage the checks they are or want to work on.

As there is no interface and it's initially just going to be done on the
Resolve credit page, we handle the form using a view that will just
redirect back to the resolve_check view.

The handling of GET requests is to get round an issue where django will
redirect a user after logging in to the assign view page with a GET
request if they click the add/remove check button, but their session has
expired.

### DIFF
--- a/mtp_noms_ops/apps/security/forms/check.py
+++ b/mtp_noms_ops/apps/security/forms/check.py
@@ -2,6 +2,7 @@ import logging
 from functools import lru_cache
 
 from django import forms
+from django.contrib import messages
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from form_error_reporting import GARequestErrorReportingMixin
@@ -171,4 +172,51 @@ class AcceptOrRejectCheckForm(GARequestErrorReportingMixin, forms.Form):
         except RequestException:
             logger.exception(f'Check {self.object_id} could not be actioned')
             self.add_error(None, _('There was an error with your request.'))
+            return False
+
+
+class AssignCheckToUserForm(GARequestErrorReportingMixin, forms.Form):
+    assignment = forms.ChoiceField(
+        choices=[
+            ('assign', _('Assign')),
+            ('unassign', _('Unassign')),
+        ],
+        required=True
+    )
+
+    def __init__(self, object_id, request, **kwargs):
+        super().__init__(**kwargs)
+        self.object_id = object_id
+        self.request = request
+
+    @cached_property
+    def session(self):
+        return get_api_session(self.request)
+
+    def get_endpoint_path(self):
+        return f'/security/checks/{self.object_id}/'
+
+    def assign_or_unassign(self):
+        endpoint = self.get_endpoint_path()
+
+        if self.cleaned_data.get('assignment') == 'assign':
+            user_id = self.request.user.pk
+        elif self.cleaned_data.get('assignment') == 'unassign':
+            user_id = None
+
+        try:
+            self.session.patch(
+                endpoint,
+                json={
+                    'assigned_to': user_id
+                }
+            )
+            return True
+        except RequestException:
+            logger.exception(f'Check {self.object_id} could not be assigned')
+            messages.add_message(
+                self.request,
+                messages.ERROR,
+                _('There was an error with your request.')
+            )
             return False
--- a/mtp_noms_ops/apps/security/tests/test_forms.py
+++ b/mtp_noms_ops/apps/security/tests/test_forms.py
@@ -7,13 +7,15 @@ from urllib.parse import parse_qs
 
 from django import forms
 from django.test import SimpleTestCase
+from django.contrib import messages
 from django.utils.timezone import make_aware
 from django.utils.dateparse import parse_datetime
+from django.utils.translation import gettext_lazy as _
 from mtp_common.auth.test_utils import generate_tokens
 from mtp_common.test_utils import silence_logger
 import responses
 
-from security.forms.check import AcceptOrRejectCheckForm, CheckListForm
+from security.forms.check import AcceptOrRejectCheckForm, CheckListForm, AssignCheckToUserForm
 from security.forms.object_base import AmountPattern, SecurityForm
 from security.forms.object_list import (
     AmountSearchFormMixin,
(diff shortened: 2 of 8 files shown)


mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -172,3 +172,54 @@ def accept_or_reject(self):
...
+        self.object_id = object_id
+        self.request = request
+
+    @cached_property
+    def session(self):
+        return get_api_session(self.request)
+
+    def get_object(self):
+        """
+        Gets the security detail object, a sender or prisoner profile
```
ushkarev, line 191: i think this was copied from somewhere else

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -172,3 +172,54 @@ def accept_or_reject(self):
...
+    def __init__(self, object_id, request, **kwargs):
+        super().__init__(**kwargs)
+        self.object_id = object_id
+        self.request = request
+
+    @cached_property
+    def session(self):
+        return get_api_session(self.request)
+
+    def get_object(self):
```
ushkarev, line 189: actually, is this method needed? you don't call it anywhere

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -28,6 +30,44 @@ class CreditsHistoryListView(SecurityView):
     form_class = CreditsHistoryListForm
 
 
+class CheckAssignView(BaseFormView):
+    """
+    Modify assignment of check
+    """
+    form_class = AssignCheckToUserForm
+    id_kwarg_name = 'check_id'
+    object_context_key = 'check'
```
ushkarev, line 39: this isn't referenced anywhere, is it needed?

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -28,6 +30,44 @@ class CreditsHistoryListView(SecurityView):
...
+    list_url = reverse_lazy('security:check_list')
```
ushkarev, line 40: also this line too

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -28,6 +30,44 @@ class CreditsHistoryListView(SecurityView):
...
+                'object_id': self.kwargs[self.id_kwarg_name],
+            },
+        )
+        return form_kwargs
+
+    def get(self, request, *args, **kwargs):
+        return HttpResponseRedirect(self.get_success_url())
+
+    def form_valid(self, form):
+        if self.request.method == 'POST':
```
ushkarev, line 59: i don't think you need this if statement, this method should only be called by `CheckAssignView.post` which itself is only called when it's a POST request

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -28,6 +30,44 @@ class CreditsHistoryListView(SecurityView):
...
+            result = form.assign_or_unassign()
+            if not result:
+                messages.add_message(
```
ushkarev, line 62: i don't think this will work. `CheckAssignView.form_valid` is only called _if_ the form was valid so adding an error message won't allow it show up anywhere. you can, however, use `messages.error(self.request, _('Credit could not be added to your list'))`

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -172,3 +172,54 @@ def accept_or_reject(self):
...
+            self.session.patch(
+                endpoint,
+                json={
+                    'assigned_to': user_id
+                }
+            )
+            return True
+        except RequestException:
+            logger.exception(f'Check {self.object_id} could not be assigned')
+            self.add_error(None, _('There was an error with your request.'))
```
ushkarev, line 224: similarly, i don't think this will work. `AssignCheckToUserForm.assign_or_unassign` is called from the view only if the form was valid so adding a form error won't allow it show up anywhere. you can, however, use `messages.error(self.request, _('Credit could not be added to your list'))`

mtp_noms_ops/templates/security/accept_or_reject_check.html
```diff
@@ -152,38 +164,44 @@ <h3 class="heading-medium">{% trans 'Credit decisions matching debit card or pri
...
-    {% endif %}
-
-  </form>
+    <form method="post">
+      {% csrf_token %}
+      {% include 'mtp_common/forms/error-summary.html' with form=form only %}
+
+      {% if check.status == 'pending' %}
+
+        <div id=accept-or-reject-form class="form-group">
```
ushkarev, line 173: it's probably better to quote attributes. is this id referenced anywhere?

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -172,3 +172,54 @@ def accept_or_reject(self):
...
+    def assign_or_unassign(self):
+        endpoint = self.get_endpoint_path()
+
+        if self.cleaned_data.get('assignment') == 'assign':
+            user_id = self.request.user.pk
+        elif self.cleaned_data.get('assignment') == 'unassign':
+            user_id = None
+        else:
+            logger.exception(f'Unknown assignment value for check {self.object_id}')
+            self.add_error(None, _('There was an error with your request.'))
```
ushkarev, line 212: see below

mtp_noms_ops/apps/security/tests/test_forms.py
```diff
@@ -2791,3 +2791,114 @@ def test_reject_with_api_error(self):
...
+
+        form = AssignCheckToUserForm(
+            object_id=check_id,
+            request=self.request,
+            data={
+                'assignment': 'foo',
+            },
+        )
+
+        self.assertTrue(form.is_valid())
```
ushkarev, line 2876: it would make sense for the form to not be valid at this stage

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -172,3 +172,54 @@ def accept_or_reject(self):
             logger.exception(f'Check {self.object_id} could not be actioned')
             self.add_error(None, _('There was an error with your request.'))
             return False
+
+
+class AssignCheckToUserForm(GARequestErrorReportingMixin, forms.Form):
+    assignment = forms.CharField(max_length=10)
```
ushkarev, line 178: you could set choices on this field to automatically disallow anything but 'assign' and 'unassign'. automatic validation is easier to deal with and shows you the options up-front

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -172,3 +172,54 @@ def accept_or_reject(self):
...
+            self.add_error(None, _('This service is currently unavailable'))
+            return {}
+
+    def get_endpoint_path(self):
+        return f'/security/checks/{self.object_id}/'
+
+    def assign_or_unassign(self):
+        endpoint = self.get_endpoint_path()
+
+        if self.cleaned_data.get('assignment') == 'assign':
```
ushkarev, line 206: validation ought to really be done by the form automatically before one tries to use it to action something.

mtp_noms_ops/apps/security/forms/check.py
mattbee: Ah yes, I think it's left from when the view was trying to render a template, which we aren't doing any more.

mtp_noms_ops/apps/security/forms/check.py
mattbee: Nope, as above :)

mtp_noms_ops/apps/security/views/check.py
mattbee: Apparently not.

mtp_noms_ops/apps/security/views/check.py
mattbee: Ah yes,

mtp_noms_ops/apps/security/views/check.py
mattbee: Happy to remove it, it was influenced by the method in the `AcceptCheckView` here: https://github.com/ministryofjustice/money-to-prisoners-noms-ops/commit/afb5af9ecc0e79a59773efd9e8a65408bcb7a46f#diff-ddf134f1183c0cbee2a271878c512447R35-R38

If we remove it here should we remove it in the other `form_valid` method?

ushkarev: ah i didn't spot that before. that uses a different base class (`SecurityDetailView`) however. some form views in this app actually submit forms using a GET so have different logic. let's leave the other one for now just in case it's actually necessary there (i haven't checked!).

mtp_noms_ops/templates/security/accept_or_reject_check.html
mattbee: Ah, yes, we need to jump down to this point (or on second thoughts early up the page) so the error messages are in view, since we moved the form tag down the page.

mtp_noms_ops/apps/security/forms/check.py
mattbee: Cool - updated to use `choiceField` and got rid of the else

mtp_noms_ops/apps/security/forms/check.py
mattbee: This isn't validating, it's using the values to determine what's being sent to the API

ushkarev: if the form validates what can go in the "assignment" field, you don't need the "else add error" part at the end

mtp_noms_ops/apps/security/views/check.py
mattbee: OK, cool I'll pop the other one back in.

mtp_noms_ops/templates/security/accept_or_reject_check.html
mattbee: We've adding in the url fragment to target the form should it error now, and this id handles the page loading so error messages and form are in page view.

mtp_noms_ops/apps/security/forms/check.py
mattbee: Cool, think we've got that sorted too

mtp_noms_ops/apps/security/tests/test_forms.py
```diff
@@ -2791,3 +2791,98 @@ def test_reject_with_api_error(self):
...
+                data={
+                    'assignment': 'assign',
+                },
+            )
+
+            form.is_valid()
+
+            self.assertEqual(form.assign_or_unassign(), False)
+            self.assertDictEqual(
+                form.errors,
```
ushkarev, line 2886: while this test may pass, these errors wouldn't be displayed anywhere cos the view redirects regardless of whether the form is valid or not.

PR #492, opened 2020-07-09
MTP-1454 add to my list button visualisation on list view

Staberinde:
I'm leaving this in draft for now as I wanted both of your opinions on the approach here. I'm not convinced of it myself and I wanted to run it by you to see if you had any better ideas

### DIFF
--- a/mtp_noms_ops/apps/security/urls.py
+++ b/mtp_noms_ops/apps/security/urls.py
@@ -375,7 +375,7 @@ urlpatterns = [
         name='resolve_check',
     ),
     url(
-        r'^security/checks/(?P<check_id>\d+)/assignment/$',
+        r'^security/checks/(?P<check_id>\d+)/assignment/(?P<list>\w+)?/?$',
         fiu_security_test(views.CheckAssignView.as_view()),
         name='assign_check',
     ),
--- a/mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
+++ b/mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
@@ -1,4 +1,4 @@
-.mtp-check-list .button {
+.mtp-check-list .button, .mtp-check-list .button-secondary {
   margin-bottom: 1em;
 
   @include media(mobile) {
@@ -20,6 +20,8 @@ table.mtp-table {
     padding-top: 0;
     margin-bottom: 0;
   }
+  border-spacing: 0;
+  border-collapse: collapse;
 }
 
 table.mtp-table--review {
@@ -164,4 +166,27 @@ table.mtp-table--review {
 
 .mtp-check-header__assigned-user {
   color: $red;
+}
+
+.mtp-check-cell__actions {
+  display: flex;
+  align-items: center;
+}
+.mtp-check-cell__list-status {
+  margin-left: 1rem;
+}
+
+.mtp-check-cell__list-status--my-list {
+  display: inline-block;
+  margin-bottom: 1.6rem;
+}
+
+.mtp-check-cell__list-status--my-list::before {
+	content: "✓";
+	font-size: 1.5em;
+	color: #00703c;
+}
+
+.mtp-check-cell__list-form {
+  display: inline-block;
 }
--- a/mtp_noms_ops/templates/security/checks_list.html
+++ b/mtp_noms_ops/templates/security/checks_list.html
@@ -78,7 +78,7 @@
       </thead>
       <tbody>
         {% for check in objects %}
-        <tr class="mtp-check-row">
+        <tr id="check-row-{{ check.id }}" class="mtp-check-row">
           <td class="mtp-compact-cell">
             {% if check.needs_attention %}
               <span class="visually-hidden">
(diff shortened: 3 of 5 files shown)


mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
```diff
@@ -165,3 +167,51 @@ table.mtp-table--review {
...
   color: $red;
 }
+
+.mtp-check-cell__added-to-my-list::before {
+	content: "✓";
+	font-size: 1.5em;
+	color: #00703c;
+}
+
+.govuk-button {
```
ushkarev, line 177: you just need to use the `button` class

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
```diff
@@ -165,3 +167,51 @@ table.mtp-table--review {
...
+.govuk-button:link, .govuk-button:visited, .govuk-button:active, .govuk-button:hover {
+	color: #ffffff;
+	text-decoration: none;
+}
+
+.govuk-button--secondary, .govuk-button--secondary:link, .govuk-button--secondary:visited, .govuk-button--secondary:active, .govuk-button--secondary:hover {
+	color: #0b0c0c;
+}
+
+.govuk-button--secondary {
```
ushkarev, line 214: you can use the `button-secondary` class

mtp_noms_ops/apps/security/urls.py
```diff
@@ -375,7 +375,7 @@ def dashboard_view(request):
         name='resolve_check',
     ),
     url(
-        r'^security/checks/(?P<check_id>\d+)/assignment/$',
+        r'^security/checks/(?P<check_id>\d+)/assignment/(?P<list>\d+|)$',
```
ushkarev, line 378: i think a word would be better than an integer. you have to search around to figure out what options exist. is it just 2 urls that we need in the end? maybe you could just have "/security/checks/???/assignment/" and "/security/checks/???/assignment/list/"

ushkarev: so maybe `r'^security/checks/(?P<check_id>\d+)/assignment/(?P<list>list|)/$'` (including the trailing slash!)

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
```diff
@@ -165,3 +167,51 @@ table.mtp-table--review {
 .mtp-check-header__assigned-user {
   color: $red;
 }
+
+.mtp-check-cell__added-to-my-list::before {
+	content: "✓";
```
ushkarev, line 172: ooh this is interesting. so this creates an inline element before a td, if i read it correctly? or does it actually go inside the td but before the rest of the td's contents?
i only ask cos if it's outside the td, then it's between two table cells and it might not render consistently in different browsers… maybe?

mtp_noms_ops/templates/security/checks_list.html
```diff
@@ -109,9 +109,27 @@ <h1 class="heading-xlarge">{{ view.title }}</h1>
...
           </td>
-          <td class="print-hidden">
-            <a href="{% url 'security:resolve_check' check_id=check.id %}">{% trans 'Review' %} <span class="visually-hidden">{% trans 'credit to' %} {{ check.credit.prisoner_name }}</span></a>
+          <td>
+            <a class="govuk-button print-hidden" href="{% url 'security:resolve_check' check_id=check.id %}">{% trans 'Review' %} <span class="visually-hidden">{% trans 'credit to' %} {{ check.credit.prisoner_name }}</span></a>
           </td>
+          {% if view.get_class_name == 'CheckListView' %}
+            {% if check.assigned_to == form.request.user.pk %}
+              <td class="mtp-check-cell__added-to-my-list">
+                Added to my list
```
ushkarev, line 118: this needs a `trans` wrapper

mtp_noms_ops/templates/security/checks_list.html
```diff
@@ -109,9 +109,27 @@ <h1 class="heading-xlarge">{{ view.title }}</h1>
...
+              </td>
+            {% elif check.assigned_to != None %}
+              <td>
+                Added to {{ check.assigned_to_name }}'s List
```
ushkarev, line 122: this needs a `blocktrans` wrapper and maybe a lowercase "List"?

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
mattbee: As it's a pseudo element, it's not really in the DOM at all, although when rendered as if it is, it's still "inside" the element it's applied to, so that's OK I think

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
mattbee: As Igor says, expecially since this our app has a different :root element font size, it makes the text in the buttons too small.

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
```diff
@@ -165,3 +167,51 @@ table.mtp-table--review {
 .mtp-check-header__assigned-user {
   color: $red;
 }
+
+.mtp-check-cell__added-to-my-list::before {
```
mattbee, line 171: Should we add a `white-space: nowrap;` rule added (to match designs and make it easier to read). Maybe at least over certain screen widths

mtp_noms_ops/templates/security/checks_list.html
```diff
@@ -109,9 +109,27 @@ <h1 class="heading-xlarge">{{ view.title }}</h1>
...
+              </td>
+            {% else %}
+              <td>
+                <form action="{% url 'security:assign_check' check_id=check.id list=1 %}" method="post">
+                  {% csrf_token %}
+                  <button type="submit" name="assignment" class="govuk-button govuk-button--secondary" value="assign"> {% trans 'Add to my List' %}</button>
```
mattbee, line 128: Can we include the prisoner name or other specific reference to the Add to list button, like the Review button has `{% trans 'Review' %} <span class="visually-hidden">{% trans 'credit to' %} {{ check.credit.prisoner_name }}</span>`

Minor point for accessibility reasons, but don't see any harm doing it.

PR #494, opened 2020-07-17
Improve "add to my list" appearance

ushkarev:
Use a more GOV.UK style check mark to indicate a credit on user's list [MTP-1487](https://dsdmoj.atlassian.net/browse/MTP-1487)

Fix layout issues in table cell with buttons and status messages dealing with "Add to my list" functionality

add-to-my-list-layout-1

---

add-to-my-list-layout-2

---

Use "Added to x’s list" consistently and fix case of "My list"

Fix display of genitive names so you get "Joe Bloggs’" instead of  "Joe Bloggs's"


mtp_noms_ops/assets-static/images/added-to-my-list.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 5.292 5.292">
```
mattbee, line 1: I don't know if it's definitely never going to be reused, but is a more generic name for the SVg (like `tick.svg`) more useful, in case this pattern gets reused?

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
```diff
@@ -171,20 +164,34 @@ table.mtp-table--review {
...
-	content: "✓";
-	font-size: 1.5em;
-	color: #00703c;
+  display: inline-block;
+  width: 20px;
+  height: 20px;
+  margin-top: -5px;
+  vertical-align: middle;
+  content: '';
+  background: file-url('added-to-my-list.png') no-repeat;
```
mattbee, line 193: As mentioned on the SVG, this feels too specific, would it be easier to reuse and find it we just call it `tick.svg` or `icon-tick.svg`?

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
```diff
@@ -171,20 +164,34 @@ table.mtp-table--review {
...
+  background-image: none, file-url('added-to-my-list.svg');
```
mattbee, line 194: Why do we need to override the image value from the background shorthand property here?

mattbee:
Nothing blocking so approved this, with a few comments/questions

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
ushkarev: this is a way to make it work on older browsers. newer browsers that understand svg also understand multiple background images, whereas older ones won't understand the `background-image` directive and not use it showing just a png. you can't do `background: file-url('added-to-my-list.png'), file-url('added-to-my-list.svg') no-repeat;` because that includes both images

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
ushkarev: fair point. it's a custom colour and size compared to the original checkmark graphic i found. still, might as well generalise the name!

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
mattbee: Ah, now it makes sense, I missed that the first image was a png

PR #495, opened 2020-07-20
Add label to highlight the description of rules a security check matched

ushkarev:
…this change depends on [a data migration in the api](https://github.com/ministryofjustice/money-to-prisoners-api/pull/603)

[MTP-1488](https://dsdmoj.atlassian.net/browse/MTP-1488)

### DIFF
--- a/mtp_noms_ops/apps/security/templatetags/security.py
+++ b/mtp_noms_ops/apps/security/templatetags/security.py
@@ -88,8 +88,6 @@ def check_description(check):
     description = check['description']
     if isinstance(description, list):
         description = '. '.join(description)
-    if description == 'Credit matched no rules and was automatically accepted':
-        return ''
     elif description.startswith('Credit matched: '):
         description = description[16:]
     return format_html(
--- a/mtp_noms_ops/apps/security/templatetags/security.py
+++ b/mtp_noms_ops/apps/security/templatetags/security.py
@@ -3,9 +3,10 @@ import logging
 
 from django import template
 from django.core.urlresolvers import reverse
-from django.utils.html import escape
+from django.utils.html import escape, format_html
 from django.utils.http import urlencode
 from django.utils.safestring import mark_safe
+from django.utils.translation import gettext
 from mtp_common.utils import format_postcode
 
 from security.forms.object_list import PrisonSelectorSearchFormMixin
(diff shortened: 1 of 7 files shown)


mattbee:
My only question is would it be worth testing that the join content filter is tested and that the rules display as expected if multiple rules are met, or are we just testing Django is working at this point?

ushkarev:
the `join` is a django-provided template filter so i think testing that directly isn't particularly valuable. instead i suppose we could make a `check_description` filter that accepts a check object. then the formatting can be changed in one place if need be. i'm happy to do that

mtp_noms_ops/apps/security/templatetags/security.py
```diff
@@ -82,6 +83,22 @@ def format_resolution(resolution):
     return credit_resolutions.get(resolution, resolution)
 
 
+@register.filter
+def check_description(check):
+    description = check['description']
+    if isinstance(description, list):
+        description = '. '.join(description)
```
Staberinde, line 90: I'm not going to block on this, but it looks like this is a comma in the wireframes

mtp_noms_ops/apps/security/templatetags/security.py
```diff
@@ -82,6 +83,22 @@ def format_resolution(resolution):
...
+    if description == 'Credit matched no rules and was automatically accepted':
```
Staberinde, line 91: I'm a bit wary of matching on magic strings. Besides why would we want to exclude this?

mtp_noms_ops/apps/security/templatetags/security.py
ushkarev: the full list won't make sense without full stops, unfortunately. each description fragment is generated as a standalone sentence

mtp_noms_ops/apps/security/templatetags/security.py
ushkarev: it's a fallback in case this filter is accidentally used for a security check that was auto-accepted, i.e. one which should _not_ be displayed using this filter

Staberinde: Hmm, i'm not entirely convinced that it's worth adding this complexity if we have confidence in our data quality. Maybe add a `logger.warning(` and we can remove this code if nothing trips it after a couple of weeks? WDYT?

ushkarev: as i say, i'm ok with removing it – is that what you'd like? i don't think we'd ever notice a log message – at present there is no such place that a log could be emitted so if we did put it in, we'd have to leave it forever. i don't think that's worth it. shall i remove this part?

PR #497, opened 2020-08-06
Update eligibility note to include contracted estate

ushkarev:
Security staff in contracted prisons are permitted access so we should make that clear in the eligibility check

[MTP-1503](https://dsdmoj.atlassian.net/browse/MTP-1503) ~[MTP-1536](https://dsdmoj.atlassian.net/browse/MTP-1536)~


ushkarev:
i just copy/pasted from an email, but actually i prefer the version with the "and". let's go for that

RoseFaulkner:
I'm unable to see the actually page (obs I'm doing something wrong) but from what you wrote Matt, I also prefer you 'and' suggestion.

PR #504, opened 2020-10-14
MTP-1511 upgrade django 2.2 LTS

Staberinde:


mattbee:
I'm assuming this is likely to pass when money-to-prisoners-common 9.21.0 is published to PyPi?

ushkarev (approved):

PR #505, opened 2020-10-15
MTP 1598 Post 2 nov banner updates

mattbee:
This is similar to the cashbook changes. I used the NOVEMBER_SECOND_CHANGES_LIVE changes rather thane BANK_TRANSFERS_ENABLED which is different to the cashbook, but I think this is better - but would it be better to have consistency for when it comes to removing these changes? I think it's OK, as it'll mostly just all be ripped out fairly soon, so shouldn't be too confusing.

### DIFF
--- a/mtp_noms_ops/templates/security/policy-change-info.html
+++ b/mtp_noms_ops/templates/security/policy-change-info.html
@@ -13,36 +13,26 @@
 
   <!--[policy-change-info]-->
   <header>
-    <h1 class="heading-xlarge">{% trans 'How Nov 2nd policy changes will affect you' %}</h1>
+    <h1 class="heading-xlarge">{% trans 'Recent policy changes (Nov 2nd 2020) that may affect your work' %}</h1>
   </header>
 
   <div class="grid-row">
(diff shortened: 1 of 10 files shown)


ushkarev (approved):

ushkarev (approved):

ushkarev (dismissed):

mtp_noms_ops/templates/security/policy-change-info.html
```diff
@@ -54,6 +44,10 @@ <h2 class="heading-medium">{% trans 'What you need to do' %}</h3>
         <li>{% trans 'transferring your money from prison to a new or existing bank account' %}</li>
       </ul>
 
+      <h2 class="heading-large">{% trans 'Prisoner bank accounts' %}</h2>
+
+      <p>{% trans 'Credits will not go into a prisoner's account if it holds £900 or more.' %}</p>
```
ushkarev, line 49: you need `’` instead of `'` in the middle here… but you probs already saw circleci complain!

ushkarev (approved):

PR #506, opened 2020-11-17
MTP-1700 Maintain current page in FIU Check List

mattbee:
Jira: https://dsdmoj.atlassian.net/browse/MTP-1700

When an FIU user added a check to their list from a page other than the
first page, the user request would be actioned and the user redirected
back to the list view. However the page param was not maintained, and
the user would be sent back to page one, regardless of which page they
were currently on.

This commit addresses that by passing the current page through with the
check data to allow the redirect to maintain the current page for the
user.


Staberinde (approved):
LGTM :+1:

PR #507, opened 2020-11-26
MTP-1654 delayed capture categorisation

Staberinde:
TODO:
* [x] Add the attrs to the decision history list and add tests 
* [x] Replace hard-coded human readable string with lookup within send money

### DIFF
--- a/mtp_noms_ops/apps/security/templatetags/security.py
+++ b/mtp_noms_ops/apps/security/templatetags/security.py
@@ -9,6 +9,7 @@ from django.utils.safestring import mark_safe
 from django.utils.translation import gettext
 from mtp_common.utils import format_postcode
 
+from security.constants import CHECK_DETAIL_RENDERED_MAPPING
 from security.forms.object_list import PrisonSelectorSearchFormMixin
 from security.models import (
     credit_sources,
@@ -338,3 +339,8 @@ def extract_best_match(context, items):
         'item': best_match,
         'total_remaining': max(len(item_list)-1, 0),
     }
+
+
+@register.filter
+def get_human_readable_check_field(field):
+    return CHECK_DETAIL_RENDERED_MAPPING[field]
--- a/mtp_noms_ops/apps/security/templatetags/security.py
+++ b/mtp_noms_ops/apps/security/templatetags/security.py
@@ -343,4 +343,4 @@ def extract_best_match(context, items):
 
 @register.filter
 def get_human_readable_check_field(field):
-    return CHECK_DETAIL_RENDERED_MAPPING[field]
+    return CHECK_DETAIL_RENDERED_MAPPING.get(field, field)
--- a/mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
+++ b/mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
@@ -190,3 +190,7 @@ table.mtp-table--review {
 .mtp-check-cell__list-form {
   display: inline-block;
 }
+
+.ignore-input {
+    display: none;
+}
--- a/mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
+++ b/mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
@@ -196,7 +196,13 @@ table.mtp-table--review {
 }
 
 .mtp-check-form-spacing > div.form-group {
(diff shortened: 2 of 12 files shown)


mtp_noms_ops/apps/security/templatetags/security.py
```diff
@@ -338,3 +339,8 @@ def extract_best_match(context, items):
         'item': best_match,
         'total_remaining': max(len(item_list)-1, 0),
     }
+
+
+@register.filter
+def get_human_readable_check_field(field):
+    return CHECK_DETAIL_RENDERED_MAPPING[field]
```
ushkarev, line 346: it's safer to do `.get(field, field)` so that a new/unexpected field still renders _something_ vaguely readable on the page without error

mtp_noms_ops/assets-src/stylesheets/views/_checks.scss
```diff
@@ -190,3 +190,19 @@ table.mtp-table--review {
 .mtp-check-cell__list-form {
   display: inline-block;
 }
+
+.ignore-input {
```
ushkarev, line 194: i don't think you need this style any longer

ushkarev (approved):
only some pretty minor notes

mtp_noms_ops/templates/security/accept_or_reject_check.html
```diff
@@ -177,13 +181,53 @@ <h3 class="heading-medium">{% trans 'Credit decisions matching debit card or pri
...
+          <fieldset class="mtp-check-form-spacing">
+            <legend class="govuk-fieldset__legend govuk-fieldset__legend">{% trans 'Select all reasons for rejecting this credit:' %}</legend>
+            {% include 'mtp_common/forms/checkbox-field.html' with field=form.payment_source_paying_multiple_prisoners input_classes='form-control-3-4' %}
+            {% include 'mtp_common/forms/checkbox-field.html' with field=form.payment_source_multiple_cards input_classes='form-control-3-4'  %}
+            {% include 'mtp_common/forms/checkbox-field.html' with field=form.payment_source_linked_other_prisoners input_classes='form-control-3-4'  %}
+            {% include 'mtp_common/forms/checkbox-field.html' with field=form.payment_source_known_email input_classes='form-control-3-4' %}
+            {% include 'mtp_common/forms/checkbox-field.html' with field=form.payment_source_unidentified input_classes='form-control-3-4' %}
+            {% include 'mtp_common/forms/checkbox-field.html' with field=form.prisoner_multiple_payments_payment_sources input_classes='form-control-3-4' %}
+            <div class="multiple-choice form-group" data-target="fiu_investigation_id">
```
ushkarev, line 202: actually, there is a minor point. it's minor cos these classes will change "soon" anyway as part of the design system update.

i think these all should just be `multiple-choice`, not `multiple-choice form-group`. when they open up, they look a bit too tightly positioned. [gds](https://govuk-elements.herokuapp.com/form-elements/#form-toggle-content) don't put form-group here. i only did the briefest of checks in chrome dev tools, so there might be some other issues/side effects to your new js toggle etc so don't just trust me on this one!

Staberinde: The uniform spacing of the check-toggled text fields with respect to the other fields seems to match justin's design. https://pm-intelligence-tool.herokuapp.com/delayed_capture_pause_03/review_0 It calls for applying uniform styling via the outer `<div>` to these check-toggled text fields when in their collapsed state and the fields being templated via `checkbox-field.html`. Without this styling the check-toggled text fields are much more distantly spaced, and it looks a bit naff. 

The reason for not using two different styling classes (i.e. a separate class for these check-toggled text fields) is that I ran into a weird issue where the margins of these fields don't stack properly, so I had to override the behaviour of all `div` elements with `form-group` class that are child elements of the `mtp-check-form-spacing` class to make their `margin-bottom` zero-width to remove all stacking in order to get these elements consistent. Currently there's no way to add a custom class to the outer `div` of the `checkbox-field.html` currently, and I was reticent to change the templating to accommodate this (even though this template snippet doesn't seem to be used anywhere else that I can find). I could add a similar rule that targets all `div` elements with `multiple-class` class that are child elements of the `mtp-check-form-spacing` class, but it felt like unnecessary duplication.

ushkarev (approved):

PR #510, opened 2021-01-19
MTP-1738 Add an auto-accept rule 

Staberinde:
Ticket link: https://dsdmoj.atlassian.net/browse/MTP-1738
Relies on: https://github.com/ministryofjustice/money-to-prisoners-api/pull/626/
To discuss:
* https://dsdmoj.atlassian.net/browse/MTP-1738?focusedCommentId=132135
* What to do on auto-accept POST/PATCH failure. Do we revert the state of the check, otherwise they won't be able to make another attempt at creating an auto-accept rule

TODO:
* [x] Fix case where auto-accept rule is added for same debit-card prisoner-combos more than once without reloading the page (reload check before accepting?)
* [x] Figure out why `api_session` taken from form in `get_context_data` doesn't have valid auth credentials bound

### DIFF
--- a/mtp_noms_ops/apps/security/constants.py
+++ b/mtp_noms_ops/apps/security/constants.py
@@ -21,6 +21,7 @@ CHECK_DETAIL_RENDERED_MAPPING = dict(
 
 CHECK_DETAIL_FORM_MAPPING = {
     'decision_reason': _('Give further details (optional)'),
+    'auto_accept_reason': _('Give reason for automatically accepting'),
     'rejection_reasons': dict(
         tuple(CHECK_REJECTION_CATEGORY_TEXT_MAPPING.items()) + tuple(CHECK_REJECTION_CATEGORY_BOOLEAN_MAPPING.items())
     )
--- a/mtp_noms_ops/apps/security/constants.py
+++ b/mtp_noms_ops/apps/security/constants.py
@@ -27,5 +27,6 @@ CHECK_DETAIL_FORM_MAPPING = {
     )
 }
 
+# This is as custom-defined exception within the API service that we match against
 CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR = \
     'An existing AutoAcceptRule is present for this DebitCardSenderDetails/PrisonerProfile pair'
--- a/mtp_noms_ops/apps/security/constants.py
+++ b/mtp_noms_ops/apps/security/constants.py
@@ -27,4 +27,4 @@ CHECK_DETAIL_FORM_MAPPING = {
     )
 }
 
-CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR = 'The fields debit_card_sender_details, prisoner_profile must make a unique set.'
+CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR = 'An existing AutoAcceptRule is present for this DebitCardSenderDetails/PrisonerProfile pair'
--- a/mtp_noms_ops/apps/security/constants.py
+++ b/mtp_noms_ops/apps/security/constants.py
@@ -27,4 +27,5 @@ CHECK_DETAIL_FORM_MAPPING = {
     )
 }
 
-CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR = 'An existing AutoAcceptRule is present for this DebitCardSenderDetails/PrisonerProfile pair'
+CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR = \
+    'An existing AutoAcceptRule is present for this DebitCardSenderDetails/PrisonerProfile pair'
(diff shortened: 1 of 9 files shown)


mtp_noms_ops/templates/security/accept_or_reject_check.html
```diff
@@ -182,9 +182,23 @@ <h3 class="heading-medium">{% trans 'Credit decisions matching debit card or pri
...
+                {% blocktrans trimmed with sender_name=check.credit.sender_name prisoner_number=check.credit.prisoner_number %}
+                  Automatically accept future credits from {{ sender_name }} to {{ prisoner_number }}
+                {% endblocktrans %}
+                <br/>
+                {% trans 'They will be flagged in Decision history.' %}
+              </label>
+            </div>
+            <div class="panel panel-border-narrow js-hidden" id="auto_accept_reason">
+              <label class="form-label" for="rejection-auto_accept_reason">{% trans 'Give reason for automatically accepting' %}</label>
+              <input class="form-control" name="auto_accept_reason" type="text" id="rejection-auto_accept_reason" disabled>
```
xoen, line 200: Maybe it's a stupid question (sorry!) but any reason for this input's ID being `rejection-auto_accept_reason`, does the ID starts with `rejection-` because of some styling or Form naming convention reason maybe?

xoen (approved):
FWIW This makes sense to me 👍

mtp_noms_ops/templates/security/accept_or_reject_check.html
ushkarev: it doesn't really matter what the id is. it's not used for styling or scripting, just to link the label to the input

ushkarev (dismissed):
haven't tried playing with it live, but looks good

ushkarev (approved):

ushkarev (dismissed):

mtp_noms_ops/apps/security/constants.py
```diff
@@ -21,7 +21,11 @@
 
 CHECK_DETAIL_FORM_MAPPING = {
     'decision_reason': _('Give further details (optional)'),
+    'auto_accept_reason': _('Give reason for automatically accepting'),
     'rejection_reasons': dict(
         tuple(CHECK_REJECTION_CATEGORY_TEXT_MAPPING.items()) + tuple(CHECK_REJECTION_CATEGORY_BOOLEAN_MAPPING.items())
     )
 }
+
+CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR = \
```
ushkarev, line 30: could we leave a comment to note that this is a magic constant referencing the api?

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -113,6 +115,34 @@ def get_form_kwargs(self):
...
+    @staticmethod
+    def is_latest_state_active(auto_accept_rule):
+        return sorted(
+            auto_accept_rule['states'], key=lambda x: x['created']
+        )[-1]['active']
+
+    def is_existing_active_auto_accept_rule(self, api_session, debit_card_sender_details_id, prisoner_profile_id):
+        query_existing_auto_accept_rule = api_session.get(
+            '/security/checks/auto-accept?{}'.format(
+                urlencode((
```
ushkarev, line 127: you don't need to url-encode these if you pass them in as a dict to the `params` argument

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -113,6 +116,37 @@ def get_form_kwargs(self):
...
+                    ('prisoner_profile_id', prisoner_profile_id),
+                    ('debit_card_sender_details_id', debit_card_sender_details_id)
+                ))
+            )
+        )
+
+        payload = query_existing_auto_accept_rule.json()
+        #  unfortunately django-filters OR's filters rather than ANDing them
```
ushkarev, line 138: (i though i left this comment somewhere, so maybe this is a duplicate or the other got lost 🤷)
i think we'll need to add this OR-ing filter to the api view. it'll be useful for the auto-accept history view too

Staberinde: You mean AND-ing? (Django Filters is already set up on the API view and it's default behavior is to OR). Seems reasonable, I can't remember how to create a django-filter that filters more than one field off the top of my head, but I'm sure it's possible.

ushkarev: yes you're right, sorry i meant AND-ing. i'm imagining a new `django_filters.FilterSet` subclass or something. could even be a custom GET param that the view responds to by pre-filtering the queryset

mtp_noms_ops/apps/security/views/check.py
Staberinde: Would that then URLencode them itself, or send the arguments as `body` parameters? I'm not sure I like the idea of a GET request with `body` parameters

ushkarev: `data` is used for POST/body whereas `params` are always added as a query string

ushkarev (approved):

PR #512, opened 2021-02-08
MTP-1734 auto accept lists

Staberinde:
Jira: https://dsdmoj.atlassian.net/browse/MTP-1734

Related PRs
API: https://github.com/ministryofjustice/money-to-prisoners-api/pull/634

TODO
* [ ] Confirm that ordering works via manual testing

### DIFF
--- a/mtp_noms_ops/apps/security/forms/check.py
+++ b/mtp_noms_ops/apps/security/forms/check.py
@@ -152,6 +152,11 @@ class AutoAcceptListForm(SecurityForm):
             'offset': 0,
             'limit': 1
         }).json()['count']
+        self.initial_index = ((self.cleaned_data.get('page', 1) - 1) * self.page_size) + 1
+        self.final_index = min(
+            self.cleaned_data.get('page', 1) * self.page_size,
+            self.total_count
+        )
         return object_list
 
 
--- a/mtp_noms_ops/apps/security/forms/check.py
+++ b/mtp_noms_ops/apps/security/forms/check.py
@@ -393,8 +393,8 @@ class AcceptOrRejectCheckForm(GARequestErrorReportingMixin, forms.Form):
                         self.session.post(
                             '/security/checks/auto-accept',
                             json={
-                                'prisoner_profile': check['credit']['prisoner_profile'],
-                                'debit_card_sender_details': check['credit']['billing_address'][
+                                'prisoner_profile_id': check['credit']['prisoner_profile'],
+                                'debit_card_sender_details_id': check['credit']['billing_address'][
                                     'debit_card_sender_details'
                                 ],
                                 'states': [{
--- a/mtp_noms_ops/apps/security/forms/check.py
+++ b/mtp_noms_ops/apps/security/forms/check.py
@@ -126,6 +126,35 @@ class CreditsHistoryListForm(SecurityForm):
         return object_list
 
 
+class AutoAcceptListForm(SecurityForm):
+    """
+    List of AutoAccepts checks.
+    """
+    def __init__(self, request, **kwargs):
+        super().__init__(request, **kwargs)
+        self.my_list_count = 0
+
+    def get_object_list_endpoint_path(self):
+        return '/security/checks/auto-accept'
+
+    def get_check_list_endpoint_path(self):
+        return '/security/checks/'
+
+    def get_object_list(self):
+        """
+        Gets objects, converts datetimes found in them.
+        """
+        object_list = convert_date_fields(super().get_object_list(), include_nested=True)
+        self.my_list_count = self.session.get(self.get_check_list_endpoint_path(), params={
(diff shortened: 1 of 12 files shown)


mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -126,6 +126,40 @@ def get_object_list(self):
...
+class AutoAcceptListForm(SecurityForm):
+    """
+    List of AutoAccepts checks.
+    """
+    def __init__(self, request, **kwargs):
+        super().__init__(request, **kwargs)
+        self.my_list_count = 0
+
+    def get_object_list_endpoint_path(self):
+        return '/security/checks/auto-accept'
```
ushkarev, line 138: no biggy, but all urls in the api end in a `/` so this will redirect on every call

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -47,6 +52,32 @@ class CreditsHistoryListView(SecurityView):
...
+    template_name = 'security/auto_accept_rule_list.html'
+    form_class = AutoAcceptListForm
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['request_params'] = self.request.GET.dict()
+        return context
+
+
+class AutoAcceptRuleDetailView(SecurityDetailView):
```
ushkarev, line 69: am i right in understanding that this view is incomplete? we should not have it in the url routes nor show the "review" button being visible at this stage. otherwise users will see a non-functioning button
```suggestion
class AutoAcceptRuleDetailView(SecurityDetailView):
```
```suggestion
class AutoAcceptRuleDetailView(SecurityDetailView):
```

mtp_noms_ops/templates/security/auto_accept_rule_list.html
```diff
@@ -0,0 +1,105 @@
...
+    {% plural %}
+    {{ number }} auto accept rules
+    {% endblocktrans %}
+  </p>
+
+  <div class="mtp-results-list mtp-check-list">
+    <table class="mtp-table table-font-xsmall">
+      <thead>
+        <tr>
+          <th>{% trans 'Debit card (from)' %}</th>
```
ushkarev, line 44: could you please add `scope="col"` to these table headers – it's something that was picked up before RE accessibility but is easy to miss. don't worry about `sortable_cell`, i added scope to it in https://github.com/ministryofjustice/money-to-prisoners-common/pull/405/

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -126,6 +126,40 @@ def get_object_list(self):
         return object_list
 
 
+class AutoAcceptListForm(SecurityForm):
+    """
+    List of AutoAccepts checks.
+    """
```
ushkarev, line 132: i'm pretty sure you need the ordering field here like in other object-listing forms otherwise no params are sent to the api

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -5340,6 +5343,114 @@ def test_resolve_page_displays_error_if_assignment_collision_from_accept_reject(
...
+                )
+                self.assertIn(
+                    f'Showing 1 to {page_size} of {api_auto_accept_response_len} auto accept rules',
+                    content
+                )
+                self.assertIn(
+                    f'{api_auto_accept_response_len} auto accept rules',
+                    content
+                )
+
```
ushkarev, line 5452: might be worth having a test that checks the ordering params are passed up to the api. just something simple given "i am sorting by column X" therefore "param_x=y" is passed to the api

mtp_noms_ops/apps/security/views/check.py
ushkarev: no idea what these "Suggested change" boxes are… i didn't put them there :)

Staberinde: Yep, this is incomplete, I only added it because I needed the url to add to the templating for the review button, and that URL had to point to a view. Happy to comment this all out till the next ticket

mtp_noms_ops/apps/security/tests/test_views.py
Staberinde: If I understand you correctly, the test would be checking that the ordering parameter was passed correctly to the HTTP request to the API service, by making a similar call to noms-ops as would be made by `sorted_cell`?

ushkarev: yes that's exactly what i would suggest. right now, i don't think it does because the form only has a page field right now

ushkarev (approved):
looks alright assuming the api responds to ordering as expected

ushkarev (approved):

PR #513, opened 2021-02-25
MTP-1735 auto accept review page

Staberinde:
Jira: https://dsdmoj.atlassian.net/browse/MTP-1735
Linked PR: https://github.com/ministryofjustice/money-to-prisoners-api/pull/638

TODO:
* [x] Unite the way we handle multiple cardholder names with elsewhere in the code
* [x] Add in Jane's user error content changes

### DIFF
--- a/mtp_noms_ops/apps/security/views/check.py
+++ b/mtp_noms_ops/apps/security/views/check.py
@@ -17,7 +17,7 @@ from security.forms.check import (
     UserCheckListForm
 )
 from security.utils import convert_date_fields
-from security.views.object_base import SecurityView, SecurityDetailView
+from security.views.object_base import SecurityView, SecurityDetailView, SimpleSecurityDetailView
 
 
 class CheckListView(SecurityView):
@@ -67,7 +67,7 @@ class AutoAcceptRuleListView(SecurityView):
         return context
 
 
-class AutoAcceptRuleDetailView(SecurityDetailView):
+class AutoAcceptRuleDetailView(SimpleSecurityDetailView, FormView):
     """
     View history of all auto-accept rules
     """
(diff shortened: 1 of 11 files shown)


mtp_noms_ops/apps/security/views/check.py
```diff
@@ -66,16 +67,72 @@ def get_context_data(self, **kwargs):
...
+    form_class = AutoAcceptDetailForm
+
+    def get_form_kwargs(self):
+        return dict(super().get_form_kwargs(), request=self.request, object_id=self.kwargs[self.id_kwarg_name])
+
+    def get_object_request_params(self):
+        return {
+            'url': f'/security/checks/auto-accept/{self.kwargs[self.id_kwarg_name]}/'
+        }
+        return super().get_object_request_params()
```
ushkarev, line 88: tiny thingy: this line is unreachable (and would throw NotImplementedError)

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -66,16 +67,72 @@ def get_context_data(self, **kwargs):
...
+        # requires the detail_object
+        list_url = self.get_list_url()
+        context_data['breadcrumbs'] = self.get_breadcrumbs(list_url)
+        return context_data
+
+    def get_object_for_template(self, obj):
+        return convert_date_fields(obj, include_nested=True)
+
+    def form_valid(self, form):
+        if self.request.method == 'POST':
```
ushkarev, line 123: we don't need this guard cos `form_valid` is only called on POST and only when the form is valid

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -66,16 +67,72 @@ def get_context_data(self, **kwargs):
...
+            result = form.deactivate_auto_accept_rule()
+            if not result:
+                return self.form_invalid(form)
+
+            messages.add_message(
+                self.request,
+                messages.INFO,
+                gettext_lazy('Auto accept rule was deactivated'),
+            )
+            return HttpResponseRedirect(self.list_url)
```
ushkarev, line 133: and this could just be `return super().form_valid(form)` if you set `success_url` on the view to be the same as `list_url`

mtp_noms_ops/apps/security/views/check.py
```diff
@@ -66,16 +67,72 @@ def get_context_data(self, **kwargs):
...
+            gettext_lazy('Review auto accept of credits from'),
+            ','.join(detail_object['debit_card_sender_details']['cardholder_names']),
+            gettext_lazy('to'),
+            detail_object['prisoner_profile']['prisoner_name']
+        )
+
+    def get_context_data(self, **kwargs):
+        context_data = super().get_context_data(**kwargs)
+
+        if not self.object:
```
ushkarev, line 101: it looks to me as thought the superclass already raises a 404 when the object can't be found. on the other hand, if it encounters an error loading the object, that's when it's an empty dict so really that should look like a 500 / "This service is currently unavailable" rather than a 404

ushkarev (commented):
only some minor points

mtp_noms_ops/apps/security/views/check.py
Staberinde: :+1: That was a vim typo. will fix

ushkarev (approved):

ushkarev (approved):

PR #516, opened 2021-03-03
MTP-1537: Remove dependencies in mtp-common (11.1.0)

xoen:
Bump `mtp-common` to `11.1.0` and remove dependencies which are already
requested there. This reduce duplication and reduce the risk of clashes
between packages versions.

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1537

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,14 +2,4 @@
 
 money-to-prisoners-common~=11.1.0
 
-Django>=2.2,<2.3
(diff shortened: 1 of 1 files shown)


ushkarev (approved):

xoen:
Thanks @ushkarev - hopefully CI will be happy this time (as I'd imagine) and we can merge 🤞

PR #517, opened 2021-03-04
Switch to GOV.UK design system

ushkarev:
MTP apps used to use a combination of [govuk-template](https://github.com/alphagov/govuk_template), [govuk-elements](https://github.com/alphagov/govuk_elements) and [govuk-frontend-toolkit](https://github.com/alphagov/govuk_frontend_toolkit). These are deprecated in favour of the more modern, flexible and accessible [govuk-frontend](https://github.com/alphagov/govuk-frontend).

Switching to the new framework involves changing ~pretty much~ every template, style and javascript module because govuk-frontend uses completely different patterns. It was decided that, during this upgrade, we should minimise our use of customised MTP-specific components where possible in favour of those provided by GDS.

[MTP-1307](https://dsdmoj.atlassian.net/browse/MTP-1307)


mtp_noms_ops/apps/security/models.py
```diff
@@ -48,7 +48,7 @@ class PaymentMethod(enum.Enum):
 security_check_statuses = {
     'pending': _('To action'),
     'accepted': _('Accepted'),
-    'auto-accepted': _('Auto Accepted'),
+    'auto-accepted': _('Auto accepted'),
```
Staberinde, line 51: This doesn't actually have an functional effect, as it's only used to display the status of a past check, and it's always coerced into all-caps in a similar way to the other statuses

mtp_noms_ops/apps/security/urls.py
```diff
@@ -386,8 +378,13 @@ def dashboard_view(request):
         name='resolve_check',
     ),
     url(
-        r'^security/checks/(?P<check_id>\d+)/assignment/(?P<list>\w+)?/(?P<current_page>\d+)?/?$',
-        fiu_security_test(views.CheckAssignView.as_view()),
+        r'^security/checks/(?P<check_id>\d+)/assignment/list/(?P<page>\d+)/$',
```
Staberinde, line 381: Why is this refactor necessary for the design system upgrade?

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -5124,88 +5124,9 @@ def test_assign_view_get_request_redirects_to_resolve_check_page(self):
 
             self.assertRedirects(response, reverse('security:resolve_check', kwargs={'check_id': check_id}))
 
-    def test_assign_view_get_request_redirects_to_resolve_check_page_with_kwarg(self):
```
Staberinde, line 5127: Why are you removing tests as part of design system upgrade PR??

Staberinde (commented):
Reviewed 7/154 files. Let me know when you get round to actioning the points from our discussion earlier and i'll review the rest of the python files. For now i'll go through the static files and submit them as another review.

mtp_noms_ops/assets-src/components/security/_results-list.scss
```diff
@@ -0,0 +1,30 @@
...
+      padding-top: 2px;
+      padding-bottom: 2px;
+
+      p {
+        margin-bottom: 0;
+      }
+    }
+
+    p {
+      font-size: 14px;
```
Staberinde, line 27: Do we want to be specifying scale-invariant sizing for paragraph-sized blocks of printed text?

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -198,18 +203,30 @@ class AcceptOrRejectCheckForm(GARequestErrorReportingMixin, forms.Form):
         required=False,
         label=CHECK_DETAIL_FORM_MAPPING['decision_reason'],
     )
-    fiu_investigation_id = forms.CharField(
+    has_fiu_investigation_id = forms.BooleanField(
```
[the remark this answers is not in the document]
ushkarev: i did revert the validation of the subfields. there's a separate branch for that. however we do need these to feed into the templates. otherwise we'll have to have a separate template for each of the 4 checkboxes… that's surely worse to keep in sync..?

mtp_noms_ops/apps/prisoner_location_admin/tasks.py
```diff
@@ -71,7 +71,7 @@ def update_locations(*, user, locations, context: Context):
                 errors += format_errors(json.loads(e.content.decode()))
             except ValueError:
                 errors.append(e.content)
-    except:  # noqa: E722
+    except:  # noqa: E722,B001
```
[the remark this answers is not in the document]
ushkarev: i'm just trying to silence an error from the newer version of flake8-bugbear :) i wasn't bothering to change any logic. `B001` ≈ `E722` just from different libs.

i've also seen linters complain about `except Exception` as "too broad a catch clause"

mtp_noms_ops/apps/security/urls.py
ushkarev: because i couldn't load the page locally anymore so i couldn't test the pages properly. the old url template gives you urls like:
`/en-gb/security/checks/502/assignment//`
but then when i click, i get a 404 saying this url does not match any pattern:
`/en-gb/security/checks/502/assignment/`
note the lack of double trailing `/`. for some reason, something locally is pre-sanitising the url before it hits the django router.

if you like i can pull this out into a separate pr, but i needed to change it to be able to test the button and what's displayed after it's clicked

Staberinde (commented):
Reviewed all static files with the exception of accept_or_reject_check.html, which I'll do seperately as it has a lot of changes.

Aside from my comment, my only question is why do we have 0 byte png files in the codebase?

mtp_noms_ops/apps/security/tests/test_views.py
ushkarev: it's a duplicate. it's line-by-line identical to the one above.

mtp_noms_ops/assets-src/components/security/_results-list.scss
ushkarev: i just copied it from what we had before:
https://github.com/ministryofjustice/money-to-prisoners-noms-ops/blob/59155a98820a8f2a4d9a77520c6cd1fbe1fecdb9/mtp_noms_ops/assets-src/stylesheets/views/_security-results.scss#L58

i'm more than happy to change it to something else… just dunno what really. maybe `smaller`?

ushkarev:
@Staberinde i have no idea why github is saying that the `mtp_noms_ops/assets-static/images/sender-profile--*.png` images are 0 bytes or empty files. i just checked out this branch freshly in a different path and i can see the images are between 500 and 3000 bytes, so they must have committed correctly :/

Staberinde (approved):
:+1: as we need to get this out by monday

PR #519, opened 2021-03-05
Add server-side validation to conditional fields

ushkarev:
…if a containing checkbox is selected (i.e. present in `form.cleaned_data`), contained subfields become required so that it's not only JS that can ensure they work correctly

TODO: perhaps these conditional fields need to remove their subfields' values from `self.cleaned_data` if not selected in case javascript is not enabled. The situation being: something is entered into a subfield but the containing checkbox is not checked. With javascript, the subfields become disabled so are not sent to the backend so are effectively treated as empty. Without javascript, the subfields remain enabled so backend will treat it as though the checkbox was selected.

### DIFF
--- a/mtp_noms_ops/apps/security/forms/check.py
+++ b/mtp_noms_ops/apps/security/forms/check.py
@@ -252,8 +252,18 @@ class AcceptOrRejectCheckForm(GARequestErrorReportingMixin, forms.Form):
         label=CHECK_DETAIL_FORM_MAPPING['rejection_reasons']['prisoner_multiple_payments_payment_sources'],
     )
 
+    # conditional fields that are hidden behind checkboxes are not required by default
+    # however if the checkbox is checked they are required
+    conditional_fields = {
+        'auto_accept': ['auto_accept_reason'],
+        'has_fiu_investigation_id': ['fiu_investigation_id'],
+        'has_intelligence_report_id': ['intelligence_report_id'],
+        'has_other_reason': ['other_reason'],
+    }
+
     error_messages = {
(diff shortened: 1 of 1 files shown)


mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -360,6 +372,32 @@ def clean(self):
                 )
         return super().clean()
 
+    def _clean_conditional_fields(self):
+        # validate conditional subfields controlled by checkboxes
+        for conditional_checkbox_field, conditional_subfields in self.conditional_fields.items():
+            if not self.cleaned_data.get(conditional_checkbox_field):
```
xoen, line 378: Python question: This is equivalent to the code below right?:

```python
if conditional_checkbox_field not in self.cleaned_data:
```

xoen (approved):
LGTM 👍 

PS: Apologies if it took me so long to review!

mtp_noms_ops/apps/security/forms/check.py
ushkarev: it's not the same. `.get` also looks up the value so this line is more like:
```
if conditional_checkbox_field not in self.cleaned_data or not self.cleaned_data[conditional_checkbox_field]:
```

xoen: Ah OK, good point, that value may be in the `cleaned_data` dictionary but it may be `False` 👌

xoen (approved):
👍

ushkarev:
> this was just a rebase off `master` branch

yup that's all i did

PR #520, opened 2021-03-05
Update display of notifications

ushkarev:
…which were accidentally left off in the [GOV.UK design system updates](https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/517).
This simply wraps existing content into lists, which is what the notifications are: lists of links. This will make the markup match the semantics of what we're showing, but have no visual impact.


xoen (approved):
👍

PR #525, opened 2021-03-15
MTP-1837: FAQ (when signed out)

xoen:
When user is not signed in, the links to "Get help" no longer take them
to the "Submit feedback" page but instead they take them to an FAQ page
with questions related to login problems etc.

This FAQ page still has links to contact us.

⚠️ **NOTE**: Currently the 'Request an account' functionality is not present in `noms-ops` (see ['Request an account - intel tool' epic](https://dsdmoj.atlassian.net/browse/MTP-1812)). 
So in the interest of making this releasable the question with the link to 'request an account' take the user to the 'Submit a ticket' page with a pre-populated message.

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1837

### DIFF
--- a/mtp_noms_ops/apps/views.py
+++ b/mtp_noms_ops/apps/views.py
@@ -1,0 +1,21 @@
+from django.conf import settings
+from django.urls import reverse_lazy
+from django.utils.translation import gettext_lazy as _
+from django.views.generic import TemplateView
+
+
+class FAQView(TemplateView):
+    template_name = 'faq.html'
+    title = _('What do you need help with?')
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context['breadcrumbs_back'] = reverse_lazy('login')
+        context['cashbook_url'] = settings.CASHBOOK_URL
+        context['reset_password_url'] = reverse_lazy('reset_password')
+        # TODO: Stop-gap solution: Replace with sign-up URL once functionality is there
+        # context['sign_up_url'] = reverse_lazy('sign-up')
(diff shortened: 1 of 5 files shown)


mtp_noms_ops/apps/views.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context['breadcrumbs_back'] = reverse_lazy('login')
+        context['cashbook_url'] = settings.CASHBOOK_URL
+        context['reset_password_url'] = reverse_lazy('reset_password')
+        # TODO: Functionality not there yet.
+        # context['sign_up_url'] = reverse_lazy('sign-up')
+        context['sign_up_url'] = 'https://example.com/TODO-sign-up-url'
```
ushkarev, line 19: in the short term and if you wanted this to be releasable, you could just point this to the contact us page with a pre-filled message. for example `I want to request an account. [Please provide your name, email address and Quantum ID]`. that way, you don't depend on the sign-up page and all that'd need changing after that's done is this one link.

xoen: Great advice as always Igor, thanks! That seems sensible, I'll do so that this is not blocked and we can add the link later 👍 💯  (I was actually wondering if I should comment that Q/A out for the time being but your suggestion is probably even better)

xoen: FYI: Update code as suggested.

ushkarev (dismissed):

xoen:
FYI: @ushkarev the stop-gap solution for the 'Request an account' question is a bit ugly and the long line was "breaking" flake8 code style so I've added a `  # noqa: E501` to ignore that. 

The idea is that once the 'Request an account' functionality is there we can update the view with that URL and the template will use the right URL.

PR #531, opened 2021-04-06
MTP-1813: Request an intel tool account

xoen:
### What
Users can request an account directly. 

It uses [existing 'Account Request' functionality in `api`](https://github.com/ministryofjustice/money-to-prisoners-api/blob/8f7882917a5ccce1d0bd7a7a529e9bab523590de/mtp_api/apps/mtp_auth/urls.py#L11) with some tweaks.

⚠️ **NOTE**: As prison is optional for these account requests and as a new "Your manager's email" field needs to be added, this PR is [dependent on the corresponding `api` PR](https://github.com/ministryofjustice/money-to-prisoners-api/pull/649) to be merged.

### Links
Ticket: https://dsdmoj.atlassian.net/browse/MTP-1813
Prototype starts: https://pm-intelligence-tool.herokuapp.com/sign-in

### DIFF
--- a/mtp_noms_ops/apps/mtp_auth/forms.py
+++ b/mtp_noms_ops/apps/mtp_auth/forms.py
@@ -48,8 +48,6 @@ class SignUpForm(BaseSignUpForm, BaseTicketForm):
             self.add_error(None, _('This service is currently unavailable'))
 
     def clean(self):
-        self.cleaned_data['role'] = 'security'
-
         if self.is_valid() and self.user_already_requested_account():
             return self.submit_ticket(
                 self.request,
--- a/mtp_noms_ops/apps/mtp_auth/forms.py
+++ b/mtp_noms_ops/apps/mtp_auth/forms.py
@@ -48,9 +48,6 @@ class SignUpForm(BaseSignUpForm, BaseTicketForm):
         self.cleaned_data['role'] = 'security'
 
         if self.is_valid() and self.user_already_requested_account():
-            # Subsequent account requests sent to ZenDesk are approved by Family Services
-            self.request.session['approver'] = _('Family Services')
-
             return self.submit_ticket(
                 self.request,
                 subject=self.account_request_zendesk_subject,
--- a/mtp_noms_ops/apps/mtp_auth/forms.py
+++ b/mtp_noms_ops/apps/mtp_auth/forms.py
@@ -23,8 +23,11 @@ class AcceptRequestForm(AcceptRequestFormBase):
 
 
 class SignUpForm(BaseSignUpForm, BaseTicketForm):
-
-    account_request_zendesk_subject = _('MTP for digital team - Prisoner money intelligence - Request for new staff account')  # noqa: E501
+    account_request_zendesk_subject = (
+        'MTP for digital team - '
+        'Prisoner money intelligence - '
+        'Request for new staff account'
+    )
     zendesk_tags = ('feedback', 'mtp', 'noms-ops', 'account_request', settings.ENVIRONMENT)
 
     role = forms.CharField(label=_('Role'), initial='security')
--- a/mtp_noms_ops/apps/mtp_auth/forms.py
+++ b/mtp_noms_ops/apps/mtp_auth/forms.py
@@ -1,6 +1,18 @@
+import logging
+
 from django import forms
-from django.utils.translation import gettext_lazy as _
-from mtp_common.user_admin.forms import AcceptRequestForm as AcceptRequestFormBase
+from django.conf import settings
+from django.utils.translation import gettext as _
+from mtp_common.user_admin.forms import (
+    AcceptRequestForm as AcceptRequestFormBase,
+    SignUpForm as BaseSignUpForm,
+)
+from oauthlib.oauth2 import OAuth2Error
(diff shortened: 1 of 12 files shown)


mtp_noms_ops/apps/mtp_auth/forms.py
```diff
@@ -0,0 +1,52 @@
...
+        except (RequestException, OAuth2Error, ValueError):
+            logger.exception('Could not look up access requests')
+            self.add_error(None, _('This service is currently unavailable'))
+
+    def clean(self):
+        self.cleaned_data['role'] = 'security'
+
+        if self.is_valid() and self.user_already_requested_account():
+            # Subsequent account requests sent to ZenDesk are approved by Family Services
+            self.request.session['approver'] = _('Family Services')
```
Staberinde, line 42: So this isn't actually true; it's FIU that review Intelligence Tool requests. But i'm not quite sure why you're adding this in the first place?

mtp_noms_ops/templates/mtp_common/user_admin/sign-up-success.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'mtp_common/user_admin/sign-up-success.html' %}
+{% load i18n %}
+{% load mtp_common %}
+
+{% block whats_next_message %}
+  <p>
+    {% blocktrans trimmed with approver=approver|wrapwithtag:'strong' %}
+      Once {{ approver }} has approved your request, you’ll get an email with your sign in details.
```
Staberinde, line 8: You can hardcode approver, as it's always going to be FIU

xoen: 🤔 It's true actual account request is always approved by FIU but in Justin prototype:
1. First request confirmation message says: ["Once **a member of the Financial Investigations Unit** has approved [...]"](https://pm-intelligence-tool.herokuapp.com/request_access_confirmation)
2. Subsequent request confirmation message says: ["Once **Family Services** has approved [...]"](https://pm-intelligence-tool.herokuapp.com/request_access_already_confirmation)

I'll double-check with Justin 🤷 

PS: This is also why when we submit a ticket to ZenDesk we set the approver to "Family Services".

mtp_noms_ops/apps/mtp_auth/forms.py
```diff
@@ -0,0 +1,52 @@
+import logging
+
+from django import forms
+from django.conf import settings
+from django.utils.translation import gettext as _
```
ushkarev, line 5: be careful about importing `gettext` vs `gettext_lazy`. it's customary in django to import the lazy version as `_`, and in fact that's the one you want. not using the lazy version means that class-level attributes would be locked to the _first_ language that loads

xoen: I see what you mean about the locking of the language 🤔 

Reason why I used the non-lazy version is because I'm using this here to translate a string that goes in the session (`Family Services` as for subsequent requests the message will be slightly different)...it turns out that you can't serialise the output of the lazy version 😬 

I thought that was simple and good enough but maybe we just use a simple constant in the session and make the translation in the view, I'll change 👍

xoen: As discussed, removed and simplified this thanks to the fact success message will be always the same.

mtp_noms_ops/templates/mtp_common/user_admin/sign-up-success.html
xoen: FYI: I was following the prototype, but turned out we actually don't want to show a different message so [I simplified the code](https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/531/commits/391efe0259e2c1d7539652fc0c47196b4d098e9d) and can resolve this and the other related comment.

mtp_noms_ops/apps/mtp_auth/forms.py
xoen: FYI: [Showing always the same success message now](https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/531/commits/391efe0259e2c1d7539652fc0c47196b4d098e9d) - resolving this.

mtp_noms_ops/templates/mtp_common/user_admin/sign-up-success.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'mtp_common/user_admin/sign-up-success.html' %}
+{% load i18n %}
+{% load mtp_common %}
+
+{% block whats_next_message %}
+  <p>
+    {% blocktrans trimmed with approver=approver|wrapwithtag:'strong' %}
```
ushkarev, line 7: you can drop the `with approver=approver|wrapwithtag:'strong'` now as the param isn't used anymore

xoen: I thought I got rid of that but well spotted. Sorted: 168abff82171c5d2fa570a00ec5f0bc48115b4f6

ushkarev (approved):
looks legit 👍

xoen:
Cheers @ushkarev ☺️

PR #532, opened 2021-04-06
Fixed tests that broke due to the change to British Summer Time on 28th March

Staberinde:


xoen (approved):
LGTM 👍

ushkarev:
FYI: instead of specifying the time zone in each test, you can just use `timezone.localtime`.
that'd automatically pick up the "Europe/London" setting. the timezone is of course exceptionally unlikely to change, but this would make the code a bit simpler

PR #533, opened 2021-04-08
MTP-1821: Add a new user

xoen:
Tweaks to the `mtp-common`'s 'New User' form:
- Tweak to opening paragraph copy
- Tweak to copy of `user_admin` checkbox label

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1821


mtp_noms_ops/apps/user_admin/views.py
```diff
@@ -0,0 +1,17 @@
+from django.utils.translation import gettext_lazy as _
+from mtp_common.user_admin.views import UserCreationView as UserCreationViewBase
+
+from .forms import UserUpdateForm
+
+
+
```
Staberinde, line 7: One too many blank lines

mtp_noms_ops/apps/user_admin/views.py
```diff
@@ -0,0 +1,17 @@
+from django.utils.translation import gettext_lazy as _
+from mtp_common.user_admin.views import UserCreationView as UserCreationViewBase
+
+from .forms import UserUpdateForm
```
Staberinde, line 4: Any reason why you've used a relative import here, but absolute imports elsewhere?

mtp_noms_ops/templates/mtp_common/user_admin/update.html
```diff
@@ -1,3 +1,69 @@
...
+            <strong aria-labelledby="id_username-label">{{ form.username.value }}</strong>
+          </p>
+        {% endif %}
+
+        {% include 'mtp_common/forms/field.html' with field=form.first_name input_classes='govuk-input--width-20' only %}
+        {% include 'mtp_common/forms/field.html' with field=form.last_name input_classes='govuk-input--width-20' only %}
+        {% include 'mtp_common/forms/field.html' with field=form.email input_classes='govuk-input--width-20' only %}
+
+        {% with field=form.role %}
+          {% if field.field.choices %}
```
Staberinde, line 44: Not sure if you need this `if` statement here. In fact you can probably let the form widget render the field itself:

mtp_noms_ops/urls.py
```diff
@@ -88,6 +89,8 @@ def root_view(request):
     ),
     url(r'^', include('security.urls', namespace='security')),
 
+    # Override mtp_common.user_admin's /users/new/ view
+    url(r'^users/new/$', UserCreationView.as_view(), name='new-user'),
```
Staberinde, line 93: Hmm, I wonder whether it would be nicer to leave defining the `new-user` url down to the service, so there's no overriding needed. What do you think?

Staberinde (commented):
Just a couple of comments

mtp_noms_ops/apps/user_admin/views.py
xoen: Darn it, I moved things around a little and forgot a newline 😛

mtp_noms_ops/apps/user_admin/views.py
xoen: You mean like `from user_admin.forms import` instead of `from user_admin.forms import`?

I think I saw this in other places in the codebase and imported the same way here, it makes sense as I'm importing from within the same module, hence the `from .forms`.

mtp_noms_ops/templates/mtp_common/user_admin/update.html
xoen: I didn't try to be clever, I copied from `mtp-common` and override only what needed to change to be on the safe side.

mtp_noms_ops/urls.py
xoen: Not sure, it makes sense for `mtp-common` to provide this functionality but I had to override some of copy, e.g. the checkbox label:

> FIU This account is for an FIU user (FIU users will be able to manage other user accounts)

I think it's nice that `mtp-common` provides a good enough/generic implementation instead of forcing the service to re-implement it.

mtp_noms_ops/apps/user_admin/views.py
xoen: https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/533/commits/5fdfda9d95b9fc23d6f8cc673fa7d8ef7de51cbb should hopefully sort this out.

ushkarev (approved):

PR #537, opened 2021-04-19
Prevent repetition of labels when displaying only "further details" on actioned credits

ushkarev:
If a credit was accepted or rejected only with "further details", 2 labels were displayed. Now, the "further details" label is displayed only if the credit was rejected and other reasons had been given as well.


xoen (approved):
LGTM 👍

PR #539, opened 2021-04-21
Move security check rejection reasons to mtp-common

ushkarev:
So that the definitions/labels are shared amongst all MTP apps and therefore data from DB can be consistently understood throughout.

See related `mtp-common` PR: https://github.com/ministryofjustice/money-to-prisoners-common/pull/421

### DIFF
--- a/mtp_noms_ops/apps/security/forms/check.py
+++ b/mtp_noms_ops/apps/security/forms/check.py
@@ -7,9 +7,14 @@ from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from form_error_reporting import GARequestErrorReportingMixin
 from mtp_common.auth.api_client import get_api_session
+from mtp_common.security.checks import (
+    CHECK_REJECTION_TEXT_CATEGORY_LABELS,
+    CHECK_REJECTION_BOOL_CATEGORY_LABELS,
+    CHECK_REJECTION_CATEGORIES,
+)
 from requests.exceptions import RequestException
 
-from security.constants import CHECK_DETAIL_FORM_MAPPING, CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR
+from security.constants import CHECK_AUTO_ACCEPT_UNIQUE_CONSTRAINT_ERROR
 from security.forms.object_base import SecurityForm
 from security.utils import convert_date_fields, get_need_attention_date
 
@@ -184,11 +189,11 @@ class AcceptOrRejectCheckForm(GARequestErrorReportingMixin, forms.Form):
     """
     CheckForm for accepting or rejecting a check.
     """
-
     fiu_action = forms.CharField(max_length=10)
+
     accept_further_details = forms.CharField(
         required=False,
-        label=CHECK_DETAIL_FORM_MAPPING['decision_reason'],
+        label=_('Give further details (optional)'),
     )
     auto_accept = forms.BooleanField(
         required=False,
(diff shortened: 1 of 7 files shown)


mtp_noms_ops/apps/security/templatetags/security.py
```diff
@@ -246,6 +246,17 @@ def format_security_check_status(check):
...
 
 
+@register.filter
+def format_security_check_rejection_reasons(check):
+    if not check or not check.get('rejection_reasons'):
+        return ''
+    return format_html_join(
+        mark_safe('<br />\n'),
+        '{}',
+        map(lambda description: [description], human_readable_check_rejection_reasons(check['rejection_reasons']))
```
xoen, line 256: This is the first time I see `format_html_join()` but from what I can tell from [the Django docs](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.html.format_html_join) it's a fancy way of joining items in a list.

Maybe it's a stupid question but why the `map` to map descriptions to lists with 1 description? Is the `map` there because `format_html_join()` wants a generator?

mtp_noms_ops/templates/security/accept_or_reject_check.html
```diff
@@ -273,10 +270,7 @@ <h2 class="govuk-heading-m">{% trans 'Decision' %}</h2>
         </p>
         <p>
           <strong>{% trans 'Decision details' %}:</strong>
-          {% for reason, value in check.rejection_reasons.items %}
-            {{ reason | get_human_readable_check_field }}{% if value != True %}: {{ value }}{% endif %}
-            <br/>
-          {% endfor %}
+          {{ check|format_security_check_rejection_reasons }}
```
xoen, line 273: Neat! Question, would the following be equivalent?

```python
{{ format_security_check_rejection_reasons(check) }}
```

ushkarev: that won't work, django template syntax does not allow calling functions like that. that's jinja style

mtp_noms_ops/apps/security/templatetags/security.py
ushkarev: `human_readable_check_rejection_reasons` returns an iterable of strings (`list[str]`) but `format_html_join` needs an iterable _of iterables_ of stuff (e.g. `list[list[Any]]`).

this function effectively works by calling `'{}'.format(*item)` for each item in the last parameter. so if the last parameter is a list of strings, the `*` expands each string into characters

ushkarev: would this be clearer?

```python
(
    (description,)
    for description in human_readable_check_rejection_reasons(check['rejection_reasons'])
)
```

mtp_noms_ops/templates/security/accept_or_reject_check.html
xoen: Cool, thanks, was just curious 👌

mtp_noms_ops/apps/security/forms/check.py
```diff
@@ -184,11 +189,11 @@ class AcceptOrRejectCheckForm(GARequestErrorReportingMixin, forms.Form):
     """
     CheckForm for accepting or rejecting a check.
     """
-
     fiu_action = forms.CharField(max_length=10)
+
     accept_further_details = forms.CharField(
         required=False,
-        label=CHECK_DETAIL_FORM_MAPPING['decision_reason'],
+        label=_('Give further details (optional)'),
```
xoen, line 196: This is nice, one less level of indirection/confusion 👍

xoen (approved):
LGTM, all seems clearer 👍

mtp_noms_ops/apps/security/forms/check.py
ushkarev: some labels weren't used anywhere else; i.e. not looked up dynamically based on variables nor from api responses. it seemed clearer to separate out the rejection reasons (which need to be dynamically looked up based on api responses) from static form labels

PR #541, opened 2021-05-07
Drop support for Python < 3.8

xoen:
We're upgrading `base`/`base-web` images to be based on Ubuntu 20.04 which
comes with Python 3.8.
It makes sense to drop support for anything older.

Also updated GH Workflow to run on newer Ubuntu 20.04 container.

Ticket: https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=79&projectKey=MTP&modal=detail&selectedIssue=MTP-1841

### DIFF
--- a/run.py
+++ b/run.py
@@ -3,8 +3,8 @@ if __name__ == '__main__':
     import os
     import sys
 
-    if sys.version_info[0:2] < (3, 6):
-        raise SystemExit('Python 3.6+ is required')
+    if sys.version_info[0:2] < (3, 8):
+        raise SystemExit('Python 3.8+ is required')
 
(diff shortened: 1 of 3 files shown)


ushkarev (approved):

run.py
```diff
@@ -4,7 +4,7 @@
     import sys
 
     if sys.version_info[0:2] < (3, 6):
```
ushkarev, line 6: ` if sys.version_info[0:2] < (3, 8):`

xoen: Yep, missed that for some reason, sorted now.

ushkarev (approved):

PR #542, opened 2021-05-07
Use timezone-aware dates in tests to match api responses

ushkarev:
Recent changes in logging levels revealed that some tests were using timezone-naive date-times.

• `mtp-api` always uses timezone-aware dates and the parsing mechanism assumes this
• tests no longer hard-code the time zone which should be determined by django settings

### DIFF
--- a/mtp_noms_ops/apps/security/tests/test_views.py
+++ b/mtp_noms_ops/apps/security/tests/test_views.py
@@ -17,8 +17,8 @@ from django.core import mail
 from django.http import QueryDict
 from django.test import SimpleTestCase, override_settings
 from django.urls import reverse
+from django.utils import timezone
 from django.utils.dateparse import parse_datetime
-from django.utils.timezone import make_aware
 from mtp_common.auth import USER_DATA_SESSION_KEY, urljoin
 from mtp_common.auth.api_client import get_request_token_url
 from mtp_common.auth.test_utils import generate_tokens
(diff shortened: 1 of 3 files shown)


mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -2677,7 +2678,7 @@ class BaseCheckViewTestCase(SecurityBaseTestCase):
     prisoner_id = 3
     credit_id = 5
 
-    credit_created_date = datetime.datetime.now()
```
ushkarev, line 2680: this was a timezone-naive date and hence usages of `credit_created_date` in mocked api responses could not be parsed properly

mtp_noms_ops/apps/security/tests/test_views.py
```diff
@@ -2830,9 +2831,7 @@ def _generate_auto_accept_response(cls, length, page_size, active=None, number_o
                             'active': not j,
                             'reason': f'I am an automatically generated auto-accept number {i}',
                             'created': (
-                                datetime.datetime.now(
-                                    tz=pytz.timezone('Europe/London')
```
ushkarev, line 2834: `'Europe/London'` shouldn't be hard-coded as there's a setting for local timezone already

xoen (approved):
Much better ❤️

xoen (approved):
👍

PR #546, opened 2021-05-27
Fix Sentry SDK initialisation

ushkarev:
and make settings configuration a bit simpler

### DIFF
--- a/mtp_noms_ops/settings/docker.py
+++ b/mtp_noms_ops/settings/docker.py
@@ -4,13 +4,10 @@ Docker settings
 from mtp_common.stack import get_current_pod
 
 from .base import *  # noqa
-from .base import ENVIRONMENT, os
+from .base import DEBUG, ENVIRONMENT, SECRET_KEY, os
 
-# SECURITY WARNING: keep the secret key used in production secret!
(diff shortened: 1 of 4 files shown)


xoen (approved):
As discussed this LGTM, it solves the sentry initialisation issue and it's a bit simpler 👍 

`DEBUG` environment variable may need to be set explicitly on local development environment as it now defaults to `False` (which is the right thing security-wise) - but unless I'm missing something it shouldn't be a problem.

mtp_noms_ops/settings/docker.py
```diff
@@ -4,13 +4,10 @@
 from mtp_common.stack import get_current_pod
 
 from .base import *  # noqa
-from .base import ENVIRONMENT, os
+from .base import DEBUG, ENVIRONMENT, SECRET_KEY, os
```
xoen, line 7: Given we have `from .base import *` above, is this import of `DEBUG` and `SECRET_KEY` for the benefit of your editor or am I missing something here?

xoen (approved):
LGTM 👍

mtp_noms_ops/settings/docker.py
ushkarev: code inspection tools don't seem to like variables that come from the `*` import 🤷

PR #551, opened 2021-08-18
Replace usage of Mailgun with GOV.UK Notify [MTP-1901]

ushkarev:


mtp_noms_ops/apps/security/tasks.py
```diff
@@ -1,37 +1,41 @@
...
-def email_export_xlsx(*, object_type, user, session, endpoint_path, filters, export_description,
-                      attachment_name):
+def email_export_xlsx(*, object_type, user, session, endpoint_path, filters, export_description):
     if not get_language():
         language = getattr(settings, 'LANGUAGE_CODE', 'en')
         activate(language)
 
     if object_type == 'credits':
-        export_message = gettext('Attached are the credits you exported from ‘%(service_name)s’.')
+        export_message = gettext('Click the link to download the credits you exported from ‘%(service_name)s’.')
```
xoen, line 23: I wonder if we could get rid of the gettext altogether here and simplify given it would always be in English anyway and the rest of the email would be in English anyway (the rest of the template itself).

But I know you probably left this as it was instead of introducing even more changes.

ushkarev: actually, i agree. i'll get rid of the gettext cos we don't want a mixed-language email to go out!

mtp_noms_ops/apps/security/tasks.py
```diff
@@ -15,26 +13,22 @@
 
 @spoolable(body_params=('user', 'session', 'filters'))
 def email_export_xlsx(*, object_type, user, session, endpoint_path, filters, export_description):
-    if not get_language():
-        language = getattr(settings, 'LANGUAGE_CODE', 'en')
-        activate(language)
-
     if object_type == 'credits':
-        export_message = gettext('Click the link to download the credits you exported from ‘%(service_name)s’.')
+        export_message = 'Click the link to download the credits you exported from ‘Prisoner money intelligence’.'
```
xoen, line 17: Bonus point for making this just a string literal 🏅

xoen (approved):
👍

xoen (approved):
LGTM 👍

PR #577, opened 2022-07-07
Fix email exporting

ushkarev:
…which called the wrong method to get filters that are passed to mtp-api. Instead, it incorrectly used the form representation.


neilmendum (approved):
LGTM

PR #591, opened 2023-05-10
MTP-1937: Remove dependency on `django-form-error-reporting` mixins

xoen:
- removed `GARequestErrorReportingMixin` from forms
- removed now unnecessary `get_ga_event_category()` method from `SearchFormV2Mixin` - this was taking advantage of an hook in `django-form-error-reporting` to customise the `category` property of the event sent to GA - it shouldn't be unnecessary now
- bumped mtp-common to version `15.0.0` (unreleased atm)

### DIFF
--- a/mtp_noms_ops/apps/security/forms/object_list.py
+++ b/mtp_noms_ops/apps/security/forms/object_list.py
@@ -37,10 +37,6 @@ class SearchFormV2Mixin(forms.Form):
     # indicates whether the form was used in advanced search
     advanced = forms.BooleanField(initial=False, required=False)
 
-    def get_ga_event_category(self):
-        """GA event category."""
-        return f'form-errors-{self.__class__.__name__}'
-
     def was_advanced_search_used(self):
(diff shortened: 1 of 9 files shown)


mtp_noms_ops/apps/security/forms/object_list.py
```diff
@@ -37,10 +37,6 @@ class SearchFormV2Mixin(forms.Form):
     # indicates whether the form was used in advanced search
     advanced = forms.BooleanField(initial=False, required=False)
 
-    def get_ga_event_category(self):
```
xoen, line 40: @ushkarev unless I'm missing something, this method was only used to override the category property of the event sent to GA right?

e.g. was overriding this: https://github.com/ministryofjustice/django-form-error-reporting/blob/e5e41f1b3e1bd6c33f6840c10ad7dadb6ebee1c3/form_error_reporting.py#L82-L86

So it should be safe (and clean) to remove - or am I missing something here?

PR #593, opened 2023-05-18
MTP-1957: Load prisoner locations from offender search

ushkarev:
…as an automated replacement for manually uploading CSV report saved from NOMIS. For now, the management command is _not_ scheduled to run periodically.

### DIFF
--- a/mtp_noms_ops/apps/prisoner_location_admin/management/commands/load_locations_from_offender_search.py
+++ b/mtp_noms_ops/apps/prisoner_location_admin/management/commands/load_locations_from_offender_search.py
@@ -1,0 +1,125 @@
+import logging
+import typing
+
+from django.conf import settings
+from django.core.management import BaseCommand
+from django.utils.functional import cached_property
+from mtp_common.auth import api_client, urljoin, MojUser
+from mtp_common.nomis import connector, request_retry
+import requests
+
+from prisoner_location_admin.models import PrisonerLocation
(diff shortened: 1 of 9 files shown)


mtp_noms_ops/apps/prisoner_location_admin/management/commands/load_locations_from_offender_search.py
```diff
@@ -0,0 +1,125 @@
...
+            while True:
+                url = urljoin(
+                    settings.HMPPS_OFFENDER_SEARCH_BASE_URL,
+                    f'/prison/{prison_id}/prisoners?cellLocationPrefix=&size=500&page={page}&sort=prisonerNumber,ASC',
+                    trailing_slash=False,
+                )
+                response: requests.Response = request_retry(
+                    'get',
+                    url,
+                    retries=0,
```
xoen, line 78: we don't want to retry here? It's a `GET` request without side effects...is it 'cause retrying could have a different set of prisoners and you prefer to rely on the retry mechanism for the whole command?

ushkarev: to be honest, _this_ call could totally retry a few times. i just wanted to trial it without any retries anywhere first so that it crashes noisily if there's any problem anywhere

xoen (dismissed):

xoen (approved):

PR #595, opened 2023-06-14
MTP-1957: Schedule loading of prisoner locations

ushkarev:


noms_ops.ini
```diff
@@ -40,3 +40,4 @@ spooler = %d/spooler
 spooler-chdir = %d
 spooler-import = mtp_%n/tasks.py
 cron = 13 8 -1 -1 1 %d/venv/bin/python %d/manage.py check_notify_templates --verbosity 2
+cron = 7 -2 -1 -1 -1 %d/venv/bin/python %d/manage.py load_locations_from_offender_search --verbosity 2
```
ushkarev, line 43: in standard cron, this is `7 */2 * * *`
i.e. 7 min past the hour, every 2 hours

mtp_noms_ops/templates/prisoner_location_admin/location_file_upload.html
```diff
@@ -35,11 +53,6 @@ <h1 class="govuk-heading-xl">{% trans 'Upload prisoner location file' %}</h1>
         </ol>
 
         <div class="govuk-inset-text">
-          <p>
-            {% blocktrans trimmed %}
-              Ensure you use the full report <strong>including the private estate</strong>.
```
ushkarev, line 40: the user running the report has no control over this, they just need to choose the right report as detailed above

noms_ops.ini
ushkarev: i wanted it to be more frequent than now (manual; once daily)… if not 2 hours, then how often would you do it? ppl do get transferred around…

xoen: Just thinking aloud really, don't have any strong opinions and I like the increased frequency and we can always reduce if it causes issues 👍

PR #598, opened 2023-10-23
Prevent scheduled prisoner location uploading outside of office hours

ushkarev:
…when _not in production_ environment because HMPPS Prisoner Search service is turned off and returns `503 Service Temporarily Unavailable`

### DIFF
--- a/mtp_noms_ops/apps/security/tests/test_forms.py
+++ b/mtp_noms_ops/apps/security/tests/test_forms.py
@@ -2835,7 +2835,7 @@ class AssignCheckToUserFormTestCase(SimpleTestCase):
             rsps.add(
                 rsps.PATCH,
                 api_url(f'/security/checks/{check_id}/'),
-                status=204,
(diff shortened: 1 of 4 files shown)


mtp_noms_ops/apps/security/tests/test_forms.py
```diff
@@ -2835,7 +2835,7 @@ def test_assign_or_unassign(self):
             rsps.add(
                 rsps.PATCH,
                 api_url(f'/security/checks/{check_id}/'),
-                status=204,
+                status=200,
```
ushkarev, line 2838: urllib3 (as of v2) now enforces content length checks… but 204 should return 0 content.
https://github.com/urllib3/urllib3/blob/main/CHANGES.rst#200-2023-04-26

as it happens, the mtp-api returns 200 on success for this endpoint anyway!

xoen (approved):

PR #608, opened 2024-02-27
MTP-1977: Redirect users back after accepting/rejecting a check.

xoen:
There are many pages of security checks to resolve. Users may be on a certain page and for whatever reason they want to be able to get back to the page they were at after accepting/rejecting a check.

The page where they were - usually the Checks list page, but it could also be in the "My list" page - is now remembered by the form. After resolving the Check the user is redirected to this page.

If for whatever reason this is missing they're taken back to the first page of the Checks list as before.

### DIFF
--- a/mtp_noms_ops/apps/security/views/check.py
+++ b/mtp_noms_ops/apps/security/views/check.py
@@ -3,6 +3,7 @@ from typing import Optional
 from django.contrib import messages
 from django.http import Http404, HttpResponseRedirect
 from django.urls import reverse, reverse_lazy
+from django.utils.http import is_safe_url
 from django.utils.translation import gettext_lazy as _
 from django.views.generic.edit import BaseFormView, FormView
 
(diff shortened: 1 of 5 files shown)


mtp_noms_ops/apps/security/views/check.py
```diff
@@ -316,6 +320,13 @@ def form_valid(self, form):
                 ui_message = _('Credit rejected')
             messages.info(self.request, ui_message)
 
-            return HttpResponseRedirect(self.list_url)
+            redirect_url = self.request.POST.get('redirect_url', self.list_url)
```
ushkarev, line 323: i think it'd be better to use the validated/pre-populated form value here?
```python
redirect_url = form.cleaned_data['redirect_url']
```

ushkarev (commented):
minor comment, otherwise ace ⭐

mtp_noms_ops/apps/security/views/check.py
xoen: 🤔 is it the same? It should. I'll give it a try, thanks for the suggestion ☺️

xoen: https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/608/commits/8e791d7a227f5ed51ed8b58b8719984f78aa64d7

xoen: It seems to work fine 👍

ushkarev (approved):

PR #615, opened 2025-02-26
Django 4.2

xoen:
Changes needed to upgrade to Django 4.x (currently 3.2 is in use).

NOTE: So far this is not using Django 4.x but these are changes required to upgrade (e.g. because something deprecated was removed)

### DIFF
--- a/mtp_noms_ops/apps/security/utils.py
+++ b/mtp_noms_ops/apps/security/utils.py
@@ -4,7 +4,7 @@ import logging
 import re
 
 from django.utils import timezone
-from django.utils.dateparse import parse_date, parse_datetime
+from django.utils.dateparse import parse_datetime
 from django.utils.translation import gettext_lazy as _
 from mtp_common.auth import USER_DATA_SESSION_KEY
 from mtp_common.auth.api_client import get_api_session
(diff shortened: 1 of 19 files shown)


mtp_noms_ops/apps/security/utils.py
```diff
@@ -38,7 +38,6 @@ def convert_date_fields(object_list, include_nested=False):
     supported yet.
     """
     fields = ('started_at', 'received_at', 'credited_at', 'refunded_at', 'created', 'triggered_at', 'actioned_at')
-    parsers = (parse_datetime, parse_date)
```
xoen, line 41: @ushkarev I've checked these fields and they all seem to be datetimes, so I've removed the logic that tried to parse `date`s.

The change to [Django 4.x' `parse_datetime()`](https://code.djangoproject.com/ticket/35097) shouldn't matter now (unless I'm missing something) and this logic _**should**_ work both on currently used Django 3.2 and on Django 4.0.

xoen: (but of course have a look and see if this makes sense to you as as well, especially as datetimes can be sneaky)

PR #617, opened 2025-03-18
Upgrade to Django 5.1 (from 3.2)

xoen:
Includes [changes to upgrade to 4.2](https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/615) and [changes required to upgrade to 5.1](https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/616).

Seems to run fine locally.

### DIFF
--- a/mtp_noms_ops/apps/security/tests/test_forms/test_search_forms.py
+++ b/mtp_noms_ops/apps/security/tests/test_forms/test_search_forms.py
@@ -462,7 +462,7 @@ class PrisonSelectorSearchFormMixinTestCase(SimpleTestCase):
         scenarios = [
             ValidationScenario(
                 {'prison_selector': PrisonSelectorSearchFormMixin.PRISON_SELECTOR_EXACT_PRISON_CHOICE_VALUE},
-                {'prison': ['This field is required']},
+                {'prison': ['This field is required.']},
             ),
             ValidationScenario(
                 {
@@ -1329,7 +1329,7 @@ class SenderFormTestCase(SecurityFormTestCase):
                     'payment_method': PaymentMethod.online.name,
                     'card_number_last_digits': '12345',
                 },
-                {'card_number_last_digits': ['You’ve entered too many characters']},
+                {'card_number_last_digits': ['Ensure this value has at most 4 characters (it has 5).']},
             ),
             ValidationScenario(
                 {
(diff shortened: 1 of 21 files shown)


mtp_noms_ops/apps/security/tests/test_forms/test_search_forms.py
```diff
@@ -1329,7 +1329,7 @@ def test_invalid(self):
                     'payment_method': PaymentMethod.online.name,
                     'card_number_last_digits': '12345',
                 },
-                {'card_number_last_digits': ['You’ve entered too many characters']},
+                {'card_number_last_digits': ['Ensure this value has at most 4 characters (it has 5).']},
```
ushkarev, line 1332: the previous message used to come from mtp-common so i don't think this is right

ushkarev: https://github.com/ministryofjustice/money-to-prisoners-common/blob/main/mtp_common/forms/__init__.py#L11-L20

ushkarev: ah i think it's cos you didn't move the app config https://github.com/ministryofjustice/money-to-prisoners-common/blob/main/mtp_common/app.py#L7

xoen: Right! Well spotted, I missed that one - I didn't realise we override some of the forms error messages.

[I've made this change in this (draft) PR](https://github.com/ministryofjustice/money-to-prisoners-common/pull/693): I'm retesting the apps locally and reverting the commits to change the form messages tests in the various apps.

mtp_noms_ops/apps/security/utils.py
```diff
@@ -38,7 +38,6 @@ def convert_date_fields(object_list, include_nested=False):
     supported yet.
     """
     fields = ('started_at', 'received_at', 'credited_at', 'refunded_at', 'created', 'triggered_at', 'actioned_at')
-    parsers = (parse_datetime, parse_date)
```
ushkarev, line 41: this change makes me quite uneasy. a date string would have a random time part invented and have the wrong type (even if it has a superset of attrs).

a smaller and less scary change would be to look for `'T'` in the string and then use `parse_datetime`, `parse_date ` otherwise

ushkarev (approved):

